PR #3, opened 2018-08-09
CDS endpoint

daco101:
Add an initial version of the CDS Hook order-review endpoint, which includes adding many CDS models (partly adapted from the swagger ones). Should respond to posts to http://localhost:8080/cds-services/coverage-requirements-discovery.

### DIFF
--- a/server/device_request.json
+++ b/server/device_request.json
@@ -1,8 +1,0 @@
-{
-  "resourceType": "DeviceRequest",
-  "note": [
-    {
-      "text": "sup dawg"
-    }
-  ]
-}
--- a/server/device_request.json
+++ b/server/device_request.json
@@ -1,0 +1,8 @@
+{
+  "resourceType": "DeviceRequest",
+  "note": [
+    {
+      "text": "sup dawg"
+    }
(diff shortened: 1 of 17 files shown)


server/device_request.json
```diff
@@ -0,0 +1,8 @@
+{
+  "resourceType": "DeviceRequest",
+  "note": [
+    {
+      "text": "sup dawg"
```
eedrummer: Please change this to something like `Stationary Compressed Gaseous Oxygen System, Rental`

server/src/main/java/endpoint/cdshooks/services/crd/CrdCdsService.java
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+
+    if (deviceRequest == null) {
+      // TODO: raise error
+    }
+
+    CdsResponse response = new CdsResponse();
+    Card card = new Card();
+    card.setSummary("im sitting on the doorstep eating bread and jam");
```
eedrummer: This should be changed as well

PR #4, opened 2018-08-13
Security

KeeyanGhoreshi:
Adds keycloak security support to the server.  Can be configured in the `application.yml file`, and the URLs that are protected can be changed in `SecurityConfig.java`.  

The security is bearer only, so there is no login page when you try to access a page without an access token in the header.  To get a token, you `POST` the authorization url `http://<HOST>/auth/realms/SpringBootKeycloak/protocol/openid-connect/token` with your credentials and it will return a bearer token.  Put that in the header of your request as `Authorization: 
bearer <TOKEN>' and it will allow access to protected endpoints.  

In order to use the authorization you need to have a keycloak server running and point `application.yml` at it.  

To run a keycloak server, download it from [here](https://www.keycloak.org/downloads.html) and unzip it, then `cd keycloak-4.2.1.Final/bin` and `./standalone.sh -Djboss.socket.binding.port-offset=100`, this is just to keep it off the same port as the fhir server.  Once it starts running you can visit it at whatever port it runs on, make an administrative account and then visit the administrative console.  Make a new realm and name it and put that name in the `application.yml`, then go to `Roles` and make a `user` role, then go to the `Users` tab near the bottom, make a user, and give them the `user` role.  

If this is getting tough to follow just follow [this tutorial](https://www.baeldung.com/spring-boot-keycloak) up until the spring boot part. 

After you've got a user, make two clients.  One should be public and have a redirect url of the Spring server, and have a `*` in the `Web Origins` category (for now).  The other client should be bearer only and whatever name it gets should be put in the `application.yml`.  

To get a token post to the token url with the following body:

`{
            grant_type:"password",
            username:"<USER>",
            password:"<PASSWORD>",
            client_id:"<THE CLIENT THAT ISN'T THE BEARER CLIENT>"
}`

Make sure the parameters are `x-www-form-urlencoded`, if you're using Postman there should be a radio button to do that.  Then use the token in the header and make a request to whatever URL is protected in `SecurityConfig.java`.  

Let me know if these steps work, if there's any questions, or bugs that occur.  I'd like to make sure everything works outside of my machine.

### DIFF
--- a/server/src/main/java/endpoint/Application.java
+++ b/server/src/main/java/endpoint/Application.java
@@ -3,7 +3,7 @@ package endpoint;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.web.servlet.ServletComponentScan;
-
+import org.springframework.boot.web.servlet.support.SpringBootServletInitializer;
 
 @SpringBootApplication
 // Finds the FhirServlet and runs it
@@ -14,9 +14,9 @@ import org.springframework.boot.web.servlet.ServletComponentScan;
  */
 public class Application {
 
+
   public static void main(String[] args) {
     SpringApplication.run(Application.class, args);
   }
-
 
 }
--- a/server/src/main/java/endpoint/controllers/DataController.java
+++ b/server/src/main/java/endpoint/controllers/DataController.java
@@ -121,17 +121,11 @@ public class DataController {
 
 
   @CrossOrigin
-  @RequestMapping(value = "/api/testing", method=RequestMethod.POST)
-  public String testingAPI(@RequestBody Object object) {
+  @RequestMapping(value = "/api/testing", method = RequestMethod.POST)
+  public String testingApi(@RequestBody Object object) {
     System.out.println(object);
     return "Thanks for posting";
   }
-
-//  @CrossOrigin
-//  @RequestMapping("/csrf")
-//  public CsrfToken csrf(CsrfToken token) {
-//    return token;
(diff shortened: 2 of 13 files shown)


server/src/main/java/endpoint/Application.java
```diff
@@ -3,7 +3,7 @@
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.web.servlet.ServletComponentScan;
-
+import org.springframework.boot.web.servlet.support.SpringBootServletInitializer;
```
eedrummer: This appears to be an unused import

server/src/main/java/endpoint/controllers/DataController.java
```diff
@@ -4,18 +4,26 @@
 import endpoint.database.DataRepository;
 
 import java.net.URI;
+import java.security.Principal;
 import java.util.Optional;
 
+import org.keycloak.representations.AccessToken;
```
eedrummer: More unused imports here

server/src/main/resources/templates/data.html
```diff
@@ -5,7 +5,9 @@
     <!-- Required meta tags -->
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-
+    <meta name="_csrf" content="${_csrf.token}"/>
```
eedrummer: It looks like we are adding CSRF tokens, but we have disabled that server side. Should we get rid of this?

PR #7, opened 2018-08-16
Major refactor for CDS Hooks

eedrummer:
Move models out of server into resources. This is to remove the case
where TestClient is dependent on the server subproject.

Updated JSON SerDe to mix FHIRContext with Jackson a little more cleanly

Slight tweak to the request-generator to fix an issue with Bundle
structure as well as the patient property on prefetch

Fixed request creator so that it now writes out a CDS Hook request
instead of FHIR Parameters

Fixed some checkstyle errors


KeeyanGhoreshi:
Everything looks fine to me

PR #12, opened 2018-08-24
Resource fetching auth

KeeyanGhoreshi:
Adds a new FHIR server with security and an auth interceptor that introspects incoming tokens.  Tokens get put in the `oauth` section of the request and the access token is passed to a client interceptor, which is how the token gets passed from incoming request to the FHIR server interceptor.  Resources from the fhir server can be fetched easily with the client using `client.read()`.  Right now the server only supports Patients but adding more resource providers is trivial.


fhir-server/src/main/java/RestfulPatientResourceProvider.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    if (thePatient.getIdentifierFirstRep().isEmpty()) {
+      /* It is also possible to pass an OperationOutcome resource
+       * to the UnprocessableEntityException if you want to return
+       * a custom populated OperationOutcome. Otherwise, a simple one
+       * is created using the string supplied below.
+       */
+      throw new UnprocessableEntityException("No identifier supplied");
+    }
+    thePatient.setId("1234");
```
plarocque4: Should this be randomly generated or an incrementing number so that a second patient created would not have the same ID?

settings.gradle
```diff
@@ -5,3 +5,5 @@ include 'operations'
 include 'server'
 include 'testingClient'
 include 'creator'
+include 'fhir-server'
```
plarocque4: Yes it's a fhir server, but perhaps it should be called provider-server?

fhir-server/src/main/java/RestfulPatientResourceProvider.java
KeeyanGhoreshi: It might be best to require the patient to have an ID when it gets created.  Forcing an incremental structure or a random generation might just make things more difficult when trying to reference specific resources.

plarocque4: That could work, I just didn't want to see a hardcoded ID that would be the same on multiple requests.

PR #19, opened 2018-09-05
Prefetch2

daco101:
This PR adds a prefetch hydrator (which includes a token resolver and associated changes), adds prefetch templates, and updates prefetch deserialization and related code.

Note: This PR breaks the CrdRequestCreator, which should be fixed in a separate task. The OrderReviewService test is ignored for now until this is changed.
Note: This PR does not use the providerFhir server for fetching prefetch elements, but should once that server supports the &_includes type queries.


resources/src/main/java/org/hl7/davinci/cdshooks/AbstractFetcherDeprecated.java
```diff
@@ -0,0 +1,71 @@
+//package org.hl7.davinci.cdshooks;
```
plarocque4: If we are no longer going to use this, should we just remove it entirely?
Are you saving it for composeClient() for now?

daco101: Yah, looks like you made more progress with the fetch authentication so my thinking was once we have our providerFhir server up your code could be used (or before I guess).

daco101: How were you testing your fetches?

plarocque4: I had tests for a utility that I added but I didn't have any unit tests for the fetches. I also hadn't gotten around to testing the round trip on the fetches due to the timing of when I implemented it.

PR #21, opened 2018-09-07
Updates to fix prefetch generation

eedrummer:
* Updated our sample request to have the correct prefetch structure
* Updated RequestBuilder to generate the correct prefetch
* Fixed serialization of OrderReviewRequest
* OrderReviewService now checks the prefetch bundle for a patient
* OrderReviewServiceTest has been re-enabled

### DIFF
--- a/resources/src/main/java/org/hl7/davinci/CrdRequestCreator.java
+++ b/resources/src/main/java/org/hl7/davinci/CrdRequestCreator.java
@@ -32,7 +32,9 @@ public class CrdRequestCreator {
     OrderReviewContext context = new OrderReviewContext();
     CrdPrefetch prefetch = new CrdPrefetch();
     request.setContext(context);
-//    request.setPrefetch(prefetch);
+    Bundle prefetchBundle = new Bundle();
+    prefetch.setDeviceRequestBundle(prefetchBundle);
+    request.setPrefetch(prefetch);
 
     DeviceRequest dr = new DeviceRequest();
     dr.setStatus(DeviceRequest.DeviceRequestStatus.DRAFT);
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/OrderReviewService.java
```diff
@@ -100,11 +100,19 @@ public CdsResponse handleRequest(@Valid @RequestBody OrderReviewRequest request)
...
+      if (patients.size() == 1) {
+        patient = patients.get(0);
+      } else {
+        // Not in the prefetch, see if it has been fetched.
+        try {
+          patient = (Patient) deviceRequest.getSubject().getResource();
+        } catch (Exception e) {
+          response
+              .addCard(CardBuilder.summaryCard("No patient could be (pre)fetched in this request"));
+        }
```
daco101: Is the aim of this change to allow us to find patients who made it into the prefetch bundle but were not appropriately linked to / referenced  in the device request?

eedrummer: To be honest, I wasn't sure how `deviceRequest.getSubject().getResource()` worked. I modified request-generator to put the Patient in the deviceRequestBundle, but it didn't seem like the code above was picking it up.

resources/src/main/java/org/hl7/davinci/CrdRequestCreator.java
```diff
@@ -53,21 +58,27 @@ public static OrderReviewRequest createRequest(Enumerations.AdministrativeGender
     patient.setGender(patientGender);
     patient.setBirthDate(patientBirthdate);
     context.setPatientId(patient.getId());
-//    prefetch.setPatient(patient);
+    bec = new Bundle.BundleEntryComponent();
+    bec.setResource(patient);
+    prefetchBundle.addEntry(bec);
     dr.setSubject(generateReference(ResourceType.Patient, patient));
```
daco101: It looks like you can use something like  `dr.setSubject(new Reference(patient));` instead of the generateReference function. This fixes the issue of the patient not getting picked up by the order review service code: `deviceRequest.getSubject().getResource();`. It looks like using the generateReference function generates a request with the same JSON output, but I think since we are passing the object into the service in the test, hapi doesn't get a chance to parse the reference string into a reference to the correct object.

PR #22, opened 2018-09-10
removed keyCloak protection

KeeyanGhoreshi:
the security still exists it just doesn't protect any endpoints.


KeeyanGhoreshi:
also I just added the ability to resize the console box on the UI.  Another small change so I just tacked it on to this PR since its inconsequential.

PR #28, opened 2018-09-12
WIP: Jwt

KeeyanGhoreshi:
Wanted to open a pull request because some of my tasks are "completed", but since they are so closely related I wanted them to all be a part of the same PR and have a platform for people to review what's been done so far. 

The request builder can now make signed JWT's and send them in the Auth header with its requests to cds-services.  The server checks the header of requests sent to endpoints specified in `SecurityConfig` and, for now, checks to make sure the header includes a structurally valid JWT.  This means it has 3 parts and includes all required keys included by the CDS-hooks specification.


request-generator/src/containers/RequestBuilder.js
```diff
@@ -10,8 +10,9 @@ import '../index.css';
 import '../components/consoleBox.css';
 import Loader from 'react-loader-spinner';
 import config from '../properties.json';
-
-
+import keypair from 'keypair';
+import base64url from "base64url";
```
eedrummer: keypair and base64url are never used and should be removed.

request-generator/src/containers/RequestBuilder.js
```diff
@@ -41,10 +41,55 @@ export default class RequestBuilder extends Component{
...
+        text.push(possible.charAt(Math.floor(Math.random() * possible.length)));
+    
+      return text.join('');
+    }
+    
+
+
+    createJwt(){
+      
+      var keypair = KEYUTIL.generateKeypair('RSA',2048);
```
eedrummer: This keypair appears to shadow the import in line 13.

server/build.gradle
```diff
@@ -39,7 +39,10 @@ dependencies {
     compile 'org.keycloak:keycloak-spring-boot-2-starter:4.0.0.Beta2'
 
     compile('com.h2database:h2')
-
+    testCompile('org.springframework.boot:spring-boot-starter-test')
+    testCompile('com.h2database:h2')
```
eedrummer: I don't think we need this line as it is included in compile.

server/src/main/java/org/hl7/davinci/endpoint/JWTAuthorizationFilter.java
```diff
@@ -0,0 +1,120 @@
...
+    super(authManager);
+  }
+  @Override
+  protected void doFilterInternal(HttpServletRequest req,
+                                  HttpServletResponse res,
+                                  FilterChain chain) throws IOException, ServletException {
+
+    String header = req.getHeader("Authorization");
+    if (header == null || !header.startsWith("Bearer")) {
+      logger.warn("JWT authorization failed - no bearer auth token present");
```
eedrummer: When I set `checkJwt` to true, I can see this log message when using Postman. However, the server still processes the request. Is this because you're still working on things or is the code intended to reject the request?

server/src/main/java/org/hl7/davinci/endpoint/JWTAuthorizationFilter.java
```diff
@@ -0,0 +1,120 @@
...
+          .keySet()
+          .containsAll(requiredPayload);
+
+      if (validHeader && validPayload) {
+        return jsonPayload.get("iss").toString();
+      } else {
+        return null;
+      }
+
+    } catch (Exception e) {
```
eedrummer: Can you make this less broad than catching `Exception`? If anything goes wrong in this method, it silently fails and returns `null`. This could make troubleshooting really hard later.

server/src/main/java/org/hl7/davinci/endpoint/JWTAuthorizationFilter.java
KeeyanGhoreshi: My update to application.yml accidentally broke the codes ability to reject requests.  I'll fix it.

KeeyanGhoreshi:
The server now verifies the signature if it can find the public key.  The code I just pushed has a new, temporary dependency to work.  The public keys exist on localhost:3000/public_keys, which doesn't have a real backend.  I'm using [json-server](https://www.npmjs.com/package/json-server) to host the keys, but once Pat's task to merge request generator and the fhir-server completes this should change.

eedrummer:
This looks good, but can you rebase on master? I think we are now out of sync with URLs with the CDS Hooks endpoints moving to /r4. I think we also likely have some conflicts between this an #31.

PR #36, opened 2018-09-18
Jwt fixes

KeeyanGhoreshi:
Fixes issues with url matching causing nonexistent endpoints to be protected, changes the format keys get sent to the server as a JWT instead of as a PEM string, adds a keystore where public keys get stored, and makes some other minor formatting changes.


server/src/main/java/org/hl7/davinci/endpoint/SecurityConfig.java
```diff
@@ -46,11 +46,11 @@ protected void configure(HttpSecurity http) throws Exception {
...
     if (myConfig.getCheckJwt()) {
-      http.csrf().disable().authorizeRequests()
+      http.authorizeRequests()
           .antMatchers().permitAll()
-          .anyRequest().authenticated();
-      http.addFilter(new JWTAuthorizationFilter(authenticationManager()))
-          .antMatcher("/cds-services/order-review-crd");
+          .anyRequest().authenticated().and()
+          .addFilter(new JwtAuthorizationFilter(authenticationManager()))
+          .antMatcher("/**/cds-services/order-review-crd");
```
eedrummer: This also needs to protect `/**/cds-services/medication-prescribe-crd`

request-generator/src/containers/RequestBuilder.js
```diff
@@ -29,13 +29,17 @@ export default class RequestBuilder extends Component{
             token: null,
             oauth:false,
             loading:false,
-            logs:[]
+            logs:[],
+            keypair:KEYUTIL.generateKeypair('RSA',2048)
```
eedrummer: It would be great if the RequestBuilder could write out the key to disk for reuse. It's probably out of scope for this PR, but something to consider. That way, the key could be shared out-of-band with the payer system. If this ever needed to take place, we would have to share the new key on every system restart

KeeyanGhoreshi:
Your build is failing because you aren't running the request-generator server that allows for REST operations on `localhost:3000/public_keys`.  The `jku` in your request is pointing at a nonexistent endpoint, and it defaults to the base url, which returns the html of the page, which produces the error you're seeing.  This is a temporary problem since Pat's tasks should create a server that hosts both the fhir server and the request generator, which will let us make the `public_keys` endpoint without using a javascript library.  

To get the project working right now, you'll need to install `json-server` with `npm install -g json-server` and then run `json-server --watch path/to/db.json`.  If you're in the request generator directory the path will be `src/db.json`.  

I'm open to other temp fixes if this won't do for `master`, like just sending the public key to the server with the request until we get things squared away.

eedrummer:
OK. That makes sense. For now, can we make the project dependent on `json-server` and have the `npm start` script run it automatically?

request-generator/src/containers/RequestBuilder.js
KeeyanGhoreshi: Are you talking about the private key?  I could just drop it into a .json file if that works.

eedrummer: I would think you need to store both.

eedrummer:
Can you rebase and resolve conflicts. There are now merge conflicts with master since I merged your checkstyle fixes.

PR #41, opened 2018-09-20
Warning card and remove devtools

eedrummer:
If the services reach the end of their processing, but the code in the
service has not added any cards, a warning card will be generated
stating that the request didn't have enough information. This typically
happens when the patient is null (broken reference) or it can't find a
CodeableConcept in the passed in request resource.

This also removes the spring devtools which was causing the server to
reload on each request.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -33,7 +33,6 @@ dependencies {
     compile('org.springframework.boot:spring-boot-starter-data-rest')
     compile('org.springframework.boot:spring-boot-starter-thymeleaf')
     compile('org.springframework.boot:spring-boot-starter-security')
-    compile("org.springframework.boot:spring-boot-devtools")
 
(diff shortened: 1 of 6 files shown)


KeeyanGhoreshi:
Everything looks fine to me, but the fact that 4 classes have to be changed with the same change to get the desired effect comes off as bad design to me.  Isn't this what an abstract class or interface is for?  I'm not all that familiar with Java and inheritance and what not but I would expect that for a situation like this the classes should extend some superclass that has the common methods.

eedrummer:
That's a valid point, but I'm not sure of a good way to handle things here. I thought that I could write a `handleRequest` method in the base `CdsService` class, but they each take different arguments that prevent that from happening. There is probably some clever way to do this with Spring AOP type stuff, but I'm not sure how to make that happen.

PR #46, opened 2018-09-21
Make services generic

KeeyanGhoreshi:
This basically overhauls the way CdsService is run.  Instead of having the base classes implement `handleRequest()`, the parent abstract class implements it.  This means that any generic operation that needs to be done only has to be added once to the `CdsService` class.  Specific methods and methods that cannot be made generic, like getting the gender of a dstu3 Patient, are offloaded to the extensions, which pass their specific information back to the parent class with abstract methods that they implement.  

These changes also require making a couple methods non-static, in order to make them available in a generic interface.  The core difference is that the requests, contexts, prefetches, and services now inherit from some abstract class or interface.  The three major changes are to `Context` and `Request`.

`Context` has a base interface which defines the shared methods that all other contexts will need, with the key element being the bundle type, which is made generic.  The interface is implemented by the `stu3` and `r4` abstract classes `ServiceContext`, which replaces the generic with the correct bundle version.  The abstract classes are extended by `medicationprescribe` and `orderreview` versions of the context.  This isn't actually needed, it just lets the method which gets the bundle get called as `getOrders` or `getMedications` respectively.

The `Request` classes inherit from `CdsRequest` and simply pass up their information, like what prefetch type they're using, which bundle version, and whether the request is coming from MedicationPrescribe or OrderReview.  `CdsRequest` fills its generics with the information passed up by the child classes.  That information is used with the generic methods defined in the abstract classes of the contexts and services to fill in the generics in `handleRequest` and complete the request.  

I did a basic test and it worked as expected.   Should probably do a bit more testing before pulling in though.  I know we're not planning on pulling this code in anytime soon, but I figured I'd put it up so people could look at it.


resources/src/main/java/org/hl7/davinci/stu3/FhirComponents.java
```diff
@@ -28,18 +29,6 @@ private FhirComponents() {
...
-   *
-   * @return the instance
-   */
-  public static FhirComponents getInstance() {
-    if (single_instance == null) {
-      single_instance = new FhirComponents();
-    }
-    return single_instance;
-  }
-
```
daco101: The FhirContext is an expensive object to make, so this code was supposed to make it so it only gets created once. Otherwise I think we create it once or twice every time we process a request.

KeeyanGhoreshi: I made the FHIR context static members of the FhirComponents class.  This should fix the problem.

eedrummer:
The git history is all crazy on this branch now. If @daco101 thinks the changes are good, I would suggest a squash and merge.

resources/src/main/java/org/hl7/davinci/stu3/FhirComponents.java
daco101: Ok, I guess that works too. Previously the whole fhircomponents object only ever had one instance but this is fine.

PR #47, opened 2018-09-21
Update read me jwt and keycloak

KeeyanGhoreshi:
Adds information about the usage of JWT and changes the keycloak documentation to reflect changes to the security landscape.


README.md
```diff
@@ -66,19 +66,22 @@ If you want to test CRD in a secure fashion using OAuth, you will need to instal
 3. Run `./standalone.sh -Djboss.socket.binding.port-offset=100` from command line to start the server.  It should run on port 8180
 4. Navigate to the KeyCloak instance in a browser, it should be at [http://localhost:8180/](http://localhost:8180)
 5. When prompted, create a new administrative username and password.
-6. Create two realms, one for the client fhir server, and one for the CRD implementation, choose any name.  Both realms follow the same following steps, the only difference should be the names of the clients
+6. Create a realm, choose any name.  The realm will be protecting the `ehr-server`.
```
eedrummer: This isn't exactly true. The realm name is hardcoded into the URL in `ClientAuthorizationInterceptor`. We should pick a realm name to use, or make the realm configurable and pull the realm name out into a properties file in ehr-server.

README.md
```diff
@@ -66,19 +66,22 @@ If you want to test CRD in a secure fashion using OAuth, you will need to instal
...
 	* Navigate to the `Users` tab and make a new user.  
 	* Give the new user a password in the `credentials` tab
 	* Go to `Role Mappings` and add the `user` role
 9. Modify config files to point at your new clients and realms
-	* Change `server/resources/application.yml` keycloak settings to point at the bearer only client in one of the realms
-	* Change `ehr-server/src/main/resources/fhirServer.properties` to have the client ID and secret of the bearer only client in the second realm
-	* Change `request-generator/src/properties.json` to include the second realm and public client ID
+	* Change `ehr-server/src/main/resources/fhirServer.properties` to have the client ID and secret of the bearer only client, then set `use_oauth` to `true` in order to use the security feature.
+		* The bearer only client's secret can be found in the `Credentials` tab
+	* Change `request-generator/src/properties.json` to include the name of the realm and the client ID of the public client
```
eedrummer: The realm name is also hardcoded into `request-generator` as well. Again, we should either make it configurable in the JSON file or just pick one and tell people to create it when setting up Keycloak.

README.md
KeeyanGhoreshi: I'll put it in both config files for requestbuilder and the ehr-server and just default the value to the realm name in the Keycloak realm I export.  That way if people import the realm it'll work out of the box.

eedrummer: I like that plan

PR #49, opened 2018-09-24
finished script to load data into ehr server

daco101:
This PR adds a gradle task to load data into the ehr server. Note that currently the ehr server db is persistent so the script doesn't need to be run constantly. Note also it will overwrite existing resources with the same type and id if they already exist.

### DIFF
--- a/ehr-server/build.gradle
+++ b/ehr-server/build.gradle
@@ -1,3 +1,13 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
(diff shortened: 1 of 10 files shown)


eedrummer:
Please update `request-generator` so that when it does not include the prefetch, the DeviceRequest it uses will properly refer to these resources.

daco101:
Ok, I think I've done that, and also fixed that null pointer exception. However, if the request-generator does not set it's own prefetch, then it will just use the patient in the ehr-server, so setting the age and gender won't work.

eedrummer:
@daco101 that makes sense. If we don't send prefetch, then the form is irrelevant.

PR #50, opened 2018-09-24
removed unnecessary files and refactored to account for removal

KeeyanGhoreshi:
removes old FHIR code, includes DaVinciMedicationRequest and DaVinciOrderRequest.


resources/src/main/java/org/hl7/davinci/stu3/FhirComponents.java
```diff
@@ -16,14 +15,6 @@
 
   private FhirComponents() {
     fhirContext = FhirContext.forDstu3();
```
eedrummer: We still need these

KeeyanGhoreshi: uh oh

PR #51, opened 2018-09-25
Changing oauth to fhirAuthorization

eedrummer:
We were using the wrong property name for a while. This has been changed
in the code, but this fixes the examples.


KeeyanGhoreshi:
Changes look fine to me

PR #54, opened 2018-09-28
Set up EHR FHIR Server so that it hosts the request-generator

plarocque4:
Set up EHR FHIR Server (ehr-server) so that it hosts the request-generator app.

### DIFF
--- a/ehr-server/README.md
+++ b/ehr-server/README.md
@@ -10,7 +10,10 @@ This subproject hosts a HAPI FHIR server that is based on the [hapi-fhir-jpaserv
 This will start the server running on http://localhost:8080/ehr-server.
 
 ## Setup the Request Generator
-See the Request Generator README.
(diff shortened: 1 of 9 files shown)


KeeyanGhoreshi:
It works for me, so I'd say its fine.

ehr-server/README.md
```diff
@@ -2,17 +2,24 @@
 This subproject hosts a HAPI FHIR server that is based on the [hapi-fhir-jpaserver-example](https://github.com/jamesagnew/hapi-fhir/tree/master/hapi-fhir-jpaserver-example).
 
 ## Adding resources to the database
-
+`gradle loadData`
```
KeeyanGhoreshi: Minor nitpick, the resources won't load until the server is running.  The way these instructions are written imply that you're supposed to run `gradle loadData` before you run `gradle appRun`, which will give you a connection error.  The instructions should probably be reordered.

PR #55, opened 2018-09-28
Public key interface

KeeyanGhoreshi:
Adds an interface to allow the addition, modification, and deletion of public keys stored on the server.  This is not for the public keys that are hosted by the EHR server (the the UI can probably just be reused there), the public keys to be displayed by this interface are keys that the server can retrieve without checking a `jku`. 

The interface is run statically like the other pages, and can be located at `localhost:8090/public`.  This means it will run when the server is run, no extra steps necessary.  

Keys can be added in any valid format that public keys can be encoded in.  I've only tested PEM and JWK.  When adding a new key, you can make up any unique ID you want, or have the program auto generate one.  The auto generated one is unique to the key, which makes it more desirable.  

You can edit the content of the key after you submit it to the server, but you can't edit its ID.  It's possible to make it so that you can edit the ID as well, but that could cause problems since the ID is supposed to be unique to a specific key.  There's no limitation on submitting two identical keys with different IDs, so if you did want to change the ID of a key, you could delete the old one and submit a new one with the same key.  

Changes made on the UI can be saved with the little save button at the top.  This just makes a POST request to the server.  The other button is a reload button.  It will query the server for the data again.   Can be helpful if developing and running into server-side troubles, as it avoids reloading the interface, or if for whatever reason the keystore data changes and the interface isn't responsible.

Keys can also be downloaded as a `json` file from the interface.


eedrummer:
Rebased and merged locally. While the code is merged in, there are a few things we should address going forward:
* The way that the app works is odd. I expected the delete button to delete the key immediately. I didn't expect to have to hit the save button at the top of the page.
* This code needs tests
* Having one big `DataController` is not great. We should refactor into two controllers.
* Clean up checkstyle violations

I'll create tasks in the JIRA backlog.

PR #57, opened 2018-10-05
React components and container for form

comeonfhqwhgads:


server/src/main/resources/templates/containers/FormContainer.jsx
```diff
@@ -0,0 +1,205 @@
...
+        <Button
+          action={this.handleFormSubmit}
+          type={"primary"}
+          title={"Submit"}
+          style={buttonStyle}
+        />{" "}
+
+        <Button
+          action={this.handleClearForm}
+          type={"secondary"}
```
KeeyanGhoreshi: This clear button will submit the form.  The `type` prop you set here isn't used in the component `Button`, which makes it a submit button by default.  Also buttons will submit forms they're in unless they have the type `button`, so that will have to be changed as well.

server/src/main/resources/templates/containers/FormContainer.jsx
```diff
@@ -0,0 +1,205 @@
...
+              value={this.state.newRule.ageLow}
+              placeholder={"Enter the low age range boundary"}
+              handleChange={this.handleInput}
+          />{" "}
+
+          <Input
+              inputType={"number"}
+              name={"ageHigh"}
+              title={"Age Range High"}
+              value={this.state.newRule.age}
```
KeeyanGhoreshi: This should be `this.state.newRule.ageHigh`, it will cause weird edge case problems that are hard to debug otherwise.

server/src/main/resources/templates/containers/FormContainer.jsx
```diff
@@ -0,0 +1,205 @@
...
+    this.setState(prevState => ({
+      newRule: { ...prevState.newRule, gender: newSelectionArray }
+    }));
+  }
+
+  handleFormSubmit(e) {
+    e.preventDefault();
+    let ruleData = this.state.newUser;
+
+    fetch("http://example.com", {
```
KeeyanGhoreshi: I'd do some validation that your state has the proper fields filled before making a request.  This isn't necessary, the server will just reject a bad request, but it would be nice to warn the user on page and not send a dead request at all.

server/src/main/resources/templates/containers/FormContainer.jsx
```diff
@@ -0,0 +1,205 @@
...
+          title={"Clear"}
+          style={buttonStyle}
+        />{" "}
+
+      </form>
+    );
+  }
+}
+
+const buttonStyle = {
```
KeeyanGhoreshi: The styling should probably be put into a `.css` file and imported into this file.  If you want to do any more styling it would start cluttering this file.

server/src/main/resources/templates/containers/FormContainer.jsx
```diff
@@ -0,0 +1,205 @@
...
+    super(props);
+
+    this.state = {
+      newRule: {
+        ruleID: "",
+        equipmentCode: "",
+        codeSystem: "",
+        ageLow: "",
+        ageHigh: "",
+        gender: [],
```
KeeyanGhoreshi: Gender being an array here is a bit misleading to what type it is.  It gets changed to a string anyway when `handleInput` is called.  It should be an empty string like the others.

server/src/main/resources/react/containers/FormContainer.jsx
```diff
@@ -0,0 +1,206 @@
...
+import TextArea from "../components/TextArea";
+import Select from "../components/Select";
+import Button from "../components/Button";
+
+class FormContainer extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      newRule: {
```
KeeyanGhoreshi: You'll need to change the key values to exactly match what the server expect (e.g. `age_range_low` instead of `ageLow`) in order for your POST to go through

server/src/main/resources/react/containers/FormContainer.jsx
```diff
@@ -0,0 +1,206 @@
...
+      newRule: { ...prevState.newRule, gender: newSelectionArray }
+    }));
+  }
+
+  //posting data to the rule's database upon Submit
+  handleFormSubmit(e) {
+    e.preventDefault();
+    let ruleData = this.state.newUser;
+
+    fetch("http://localhost:8090/data/api", {
```
KeeyanGhoreshi: this should be `http://localhost:8090/api/data`

server/src/main/resources/templates/data.html
```diff
@@ -60,6 +61,7 @@ <h1 th:text="${foo}"></h1>
             <!--</tr>-->
             <tr>
                 <td><input type="submit" value="Submit" /></td>
+                <td><input type="delete" value="Delete" /></td>
```
KeeyanGhoreshi: This doesn't add a button it adds an input with the word "Delete" in it.  It should probably be a button.  But more importantly the delete button should be embedded in the table entries so it can be easily clicked for each entry.

server/src/main/resources/react/containers/FormContainer.jsx
```diff
@@ -99,6 +100,29 @@ class FormContainer extends Component {
       });
     });
   }
+  
+  //Deleting rule from rule db
+  handleRuleDelete(rule){
+    var data = {
+        id: rule.id
+    }
+    fetch("http://localhost:8090/api/data", {
```
KeeyanGhoreshi: posting to `/api/data` will not delete the rule, it will just replace it with a rule that only has an ID.  Use the delete mapping at `/api/data/{id}` to delete the rule from the database.

server/src/main/resources/react/containers/FormContainer.jsx
```diff
@@ -99,6 +100,29 @@ class FormContainer extends Component {
...
+        method: 'POST',
+        headers: {'Content-Type': 'application/json'},
+        body: JSON.stringify(data)
+    }).then(function(response) {
+        if (response.status >= 400) {
+          throw new Error("Bad response from server");
```
KeeyanGhoreshi: This error message should probably include the status code.

server/src/main/resources/react/containers/FormContainer.jsx
```diff
@@ -106,16 +106,15 @@ class FormContainer extends Component {
     var data = {
         id: rule.id
     }
-    fetch("http://localhost:8090/api/data", {
-        method: 'POST',
-        headers: {'Content-Type': 'application/json'},
+    fetch("http://localhost:8090/api/data/{id}", {
```
KeeyanGhoreshi: the delete endpoint doesn't take a body, it just takes the url parameter.  This should be `"http://localhost:8090/api/data/"+rule.id`.

server/src/main/resources/react/containers/FormContainer.jsx
```diff
@@ -122,6 +122,16 @@ class FormContainer extends Component {
...
     });
 }
+    //allow updating of rule
+	handleRuleUpdate(rule) {
+	const request = new Request('"http://localhost:8090/api/data/"+rule.id', {
+      method: 'PUT',
+      headers: new Headers({
+        'Content-Type': 'application/json'
+      }), 
+      body: JSON.stringify({rules: rule})
```
KeeyanGhoreshi: what is {rules: rule} doing here?

server/src/main/resources/react/containers/FormContainer.jsx
```diff
@@ -124,7 +124,8 @@ class FormContainer extends Component {
     });
 }
     //allow updating of rule
-	handleRuleUpdate(rule) {
+	handleRuleUpdate() {
+	action=(()=>{handleRuleUpdate(newRule)})
```
KeeyanGhoreshi: This will throw an error since there's no function `handleRuleUpdate(newRule)`.  This type of big arrow function calling another function with an argument should be used as a callback.  For example, in the edit button, the action (or onClick) function should be `()=>{handleRuleUpdate(newRule)}`.  It should not be inside the a function like this.  

Regardless, it doesn't look like `newRule` is coming from anywhere.  I'd suggest leaving the function without an argument (`handleRulesUpdate()`) and having it access the rule that's stored in the state, since that should be the new rule that should be getting submitted.

server/src/main/resources/react/src/containers/FormContainer.jsx
```diff
@@ -228,18 +227,59 @@ class FormContainer extends Component {
...
+          title={"Edit"}
           style={buttonStyle}
         />{" "}
       </form>
     );
   }
 }
 
+//Trying to add row components to a table where users can edit and delete individually -- should this be part of the above render?
+class Row extends Component{
```
KeeyanGhoreshi: Generally speaking we should keep it to 1 class per file.  Things get more difficult to track when a file has multiple classes.

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -27,8 +33,234 @@ export default class DataTable extends Component {
...
+    };
+    this.handleTextArea = this.handleTextArea.bind(this);
+    this.handleFormSubmit = this.handleFormSubmit.bind(this);
+    this.handleClearForm = this.handleClearForm.bind(this);
+    this.handleCheckBox = this.handleCheckBox.bind(this);
+    this.handleInput = this.handleInput.bind(this);
+    this.handleRuleDelete = this.handleRuleDelete.bind(this);
+    this.handleRuleUpdate = this.handleRuleUpdate.bind(this);
+    this.handleSelect = this.handleSelect.bind(this);
+  }
```
KeeyanGhoreshi: This entire section is under the `componentDidMount()` lifecycle function.  All this code, with the exception of the code that was already there, needs to move out of this section, with the `constructor` stuff, like initializing the state and binding the functions to `this`, moving into the constructor of this file and the function definitions moving out into the top level.

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -1,4 +1,10 @@
 import React, {Component} from 'react';
+import CheckBox from "../components/CheckBox";
```
KeeyanGhoreshi: You will need to make a directory in `node_interfaces/src/components` for the components.  They will not load as the code is now.  For example, from this file you might import with `import CheckBox from './TableForm/Checkbox';

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -1,4 +1,10 @@
...
+import Input from "../components/Input";
+import TextArea from "../components/TextArea";
+import Select from "../components/Select";
+import Button from "../components/Button";
```
KeeyanGhoreshi: Also, on the topic of imports, the data table really shouldn't integrate all the code for the form.  There should be a `Form` component which `DataTable` imports.

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -62,6 +294,8 @@ export default class DataTable extends Component {
                                 <td >{rule.providerAddressState}</td>
                                 <td >{rule.noAuthNeeded?"false":"true"}</td>
                                 <td >{rule.infoLink}</td>
+                                <td className="glyphicon glyphicon-trash" onClick={() => {this.props.handleRuleDelete(this.props.row)}}></td>
```
KeeyanGhoreshi: The function can be called with `this.handleRuleDelete(rule)`, it will throw an error due to the inclusion of `props`.  The `props` only contain things passed into the component when it is instantiated in the JSX.  In this case nothing is passed into the props.

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -62,6 +294,8 @@ export default class DataTable extends Component {
...
+                                <td className="glyphicon glyphicon-edit" onClick={() => {this.props.handleRuleEdit(this.props.row)}} ></td>
```
KeeyanGhoreshi: Same thing here, and also `this.props.row` will return undefined, which will cause an error.

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -27,8 +33,234 @@ export default class DataTable extends Component {
...
+          gender: [],
+          infoLink: "",
+          noAuthNeeded: ""
+      }
+    });
+  }
+
+
+  //Displaying the form fields and handling their inputs
+  render() {
```
KeeyanGhoreshi: There should only be one `render` function per file.  This form info should be put in a different component file and imported to the data table.

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -27,8 +33,234 @@ export default class DataTable extends Component {
...
+          style={buttonStyle}
+        />{" "}
+
+        <Button
+          action={this.handleClearForm}
+          type={"secondary"}
+          title={"Clear"}
+          style={buttonStyle}
+        />{" "}
+      </form>
```
KeeyanGhoreshi: Also, keep in mind that the home page (`/`) and the admin interface (`/data`) both use this `DataTable` class.  Adding the form directly here would add the form onto the home page, which we don't want.  This can be handled with a `prop`, passed in from `HomePage` indicating that the form does not need to be loaded.

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -27,8 +33,234 @@ export default class DataTable extends Component {
...
+
+      const buttonStyle = {
```
KeeyanGhoreshi: This should be put into `index.css`

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -72,7 +306,4 @@ export default class DataTable extends Component {
              </div>
 
          )
-
-    }
-}
```
KeeyanGhoreshi: There might be a missing brace here.

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -27,8 +33,234 @@ export default class DataTable extends Component {
...
+          }
+
+           //Deleting rule from rule db
+          handleRuleDelete(rule){
+            var data = {
+                id: rule.id
+            }
+            fetch('"http://localhost:8090/api/data/"+rule.id', {
+                method: 'DELETE',
+                body: JSON.stringify(rule)
```
KeeyanGhoreshi: The `DELETE` method does not need a body.

server/src/main/resources/node_interfaces/src/components/index.css
```diff
@@ -0,0 +1,28 @@
...
+style={buttonStyle}
+/>{" "}
+
+<Button
+action={this.handleRuleUpdate}
+handleChange={this.state.newRule}
+type={"secondary"}
+title={"Edit"}
+style={buttonStyle}
+/>{" "}
```
KeeyanGhoreshi: This should not be in `index.css`

server/src/main/resources/node_interfaces/src/components/DataTable.js
```diff
@@ -16,251 +11,6 @@ export default class DataTable extends Component {
          
     }
 
-
-    componentDidMount(){
-
-        document.body.style.backgroundColor = "white"// Set the style
-
-        const data = fetch('http://localhost:8090/api/data', {
```
KeeyanGhoreshi: This part of the data table (where the data gets fetched from the server) should remain here.  Unless you plan on fetching the data somewhere else and passing it to the data table, which would be fine too.

server/src/main/resources/node_interfaces/src/components/index.css
```diff
@@ -1,28 +1,3 @@
...
-action={this.handleRuleUpdate}
-handleChange={this.state.newRule}
-type={"secondary"}
-title={"Edit"}
-style={buttonStyle}
-/>{" "}
\ No newline at end of file
+const buttonStyle = {
+  margin: "10px 10px 10px 10px"
+};
```
KeeyanGhoreshi: This needs to be changed to be CSS (it's js right now).  Should look like this:
`.buttonStyle {
      margin: 10px 10px 10px 10px;
}`

server/src/main/resources/node_interfaces/src/components/Form.js
```diff
@@ -213,6 +213,35 @@ class Form extends Component {
           handleChange={this.handleTextArea}
           placeholder={"Enter link to further information"}
         />
+
+        <Button
+          action={this.handleFormSubmit}
+          type={"primary"}
+          title={"Submit"}
+          style={buttonStyle}
```
KeeyanGhoreshi: This will cause an error since `buttonStyle` is now a css class and not a js object.  To apply the style you'll need to replace this with `className="buttonStyle"`.

KeeyanGhoreshi:
The recent changes should address all the outstanding issues.  I changed the request log and key interface to have light themes to match the rules table.  Changing them back to a dark theme would be trivial, and changing the rules table to a dark theme to match would also be relatively simple using an inverting filter since everything is grayscale.  So if we want to switch styles later it can be done.  

Otherwise, to update rules simply click on the little edit button at the far end of the table and all the columns (except the id) of that row will be replaced with either a text input or a dropdown where you can enter new values.  To finish editing simply hit `enter` in a text input or click the edit button again.  

The little trashcan will delete the row it belongs to.  There is currently no prompt to ensure the user meant to click it.  The plus sign button at the top will add a new entry.

The only oddity of these changes is the handling of the ID.  Currently the server handles giving new rules an ID, and will override any ID that gets provided.  The server will respond to a post request with the ID of the new rule, which is then used by the rules table interface to correctly label the rule.  This behavior is a bit counterintuitive, I'd argue the server shouldn't be changing explicitly provided ID numbers and instead reject requests with an ID number that overlaps with an existing rule and continue to generate an ID for rules that get sent with a `null` ID or lacking the element completely.  

In any case, the  rules table can be run by running the `node_interfaces` folder with `npm start`.  It is not yet built out into the static folder, so it will not appear on the server's version of the node interfaces.  

Another note, the format of the rules is currently all over the place in the code.  It would be good to have a task that consolidates the rule format somewhere.

PR #58, opened 2018-10-09
Public key fixup

KeeyanGhoreshi:
Takes care of issues with public key interface (minus tests which are being worked on, since this update makes large changes to the way public key interface works). 

Instead of a JSON file public keys now have their own JPA repository.  Using JSON was causing problems with deletion and editing of keys, and the tests couldn't harmlessly write to the file.  Keys are now saved with an `id` string, their `kid`, and a `key`, which is their content.  The content can be any string.


server/src/main/java/org/hl7/davinci/endpoint/SigningKeyResolverCrd.java
```diff
@@ -26,6 +31,7 @@
 import java.security.spec.RSAPublicKeySpec;
 import java.util.Base64;
 import java.util.Objects;
+import java.util.Optional;
```
eedrummer: This is an unused import. There are a lot of these that should be cleaned up.

server/src/main/java/org/hl7/davinci/endpoint/controllers/PublicKeyController.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.io.FileReader;
```
eedrummer: It appears that all of the java.io.* imports are unused.

server/src/test/java/PublicKeyControllerTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @LocalServerPort
+  private int port;
+
+  @Autowired
+  private TestRestTemplate restTemplate;
+
+  @Before
+  @Test
+  public void testPostKey() {
```
eedrummer: This isn't a particularly useful test. Other than postForObject throwing and exception, would there be any way to tell if the post functionality of the controller broke?

KeeyanGhoreshi: If it 500s or 404s the test will fail.  But if the post functionality broke silently than the next test which checks the repository for the posted key should fail.

eedrummer: Also, does that mean that the tests are order dependent? Some test frameworks will change the order of tests to make sure there aren't hidden dependencies. I know that MiniTest in ruby does that, JUnit may be different.

KeeyanGhoreshi: JUnit has a non-determinable ordering but the `@Before` should make sure that the `post` is run before the `get`.

PR #63, opened 2018-10-18
Add address state

daco101:
This PR adds support for patient and practitioner state. Note that the practitioner state is not currently set up for stu3 at this time. I also refactors the services class, and a few other refactors.


eedrummer:
Rebased and merged locally with a fix for DropDown state.

PR #69, opened 2018-11-05
created basic smart on FHIR app

KeeyanGhoreshi:
This is a basic express app that hosts the launch/index files needed for a bare-bones SMART app.  It runs on `localhost:3050` and can be run with `npm start`.  It will show the name of the patient that is currently in context ( it will also show "Medication List" but that will be empty).  

Launch it [here](https://launch.smarthealthit.org/) using the url http://localhost:3050/launch


eedrummer:
Can we close this and put the code in the `dtr` repo?

PR #73, opened 2018-11-06
Ui consolidation

KeeyanGhoreshi:
This pulls all the UI components into a single node project.  The Spring Controller now exclusively points at the `index.html` file, and routing is handled by React.  The Thymeleaf pages have been replaced with React components.  The commit `introduced routing to the react app` contains the majority of the actual changes that aren't moving files around, deleting old files, or changing path names.

Future changes to the data table or home page should be done in the new react files.  The form is absent from the administrative data interface and will hopefully be replaced with the app that Jess is building.


eedrummer:
It's awesome how quickly you were able to get this together. That said, I'm having trouble getting the public key interface working locally.

When I make a request with crd-request-generator, it looks like it is passing a JWT. However, when I hit  `http://localhost:8090/api/public`, it only returns an empty array. I think that because of that, I'm not seeing anything in the keys page. I think that either the saving of JWTs is not working or the API isn't properly returning them.

KeeyanGhoreshi:
Can you make sure that the `checkJwt` option in `application.yml` is `true`?  It defaults to `false` and circumvents the security pipeline which is where the key would be checked into the key repository.

eedrummer:
@KeeyanGhoreshi you're right. I thought it defaulted to true. Thanks for setting me straight.

PR #74, opened 2018-11-15
Smart app cards

KeeyanGhoreshi:
This PR adds a link to the cards returned to the request generator which launches the smart on FHIR app.  The DTR app needs to be running for it to work.  To test it you can switch the ID of the patient on the incoming request from the request generator (or directly make a request to one of the service endpoints and follow the link returned in the JSON and choose some patient ID) to some patient that exists in the EHR server.  The program will fail currently if the patient does not have a name, which most of the sample patients do not have. 

The app launch will trigger a chain of auth events, first acquiring the token and authorization keycloak endpoints from the fhir server conformance statement.  The token and authorization endpoints can be set in `fhirServer.properties` in the resources folder of the EHR server.  They will default to the keycloak server that the project used to use.  

The launch page gets the conformance statement, parses out the token and authorization endpoints, and acquires an authorization code from the auth endpoint.  It will redirect you to the login page and you'll have to login using some user account that's set up on your local running keycloak instance.  This behavior can be swapped out for some token based approach, as the code  received from logging in gets saved in a cookie and is used to bypass this step on any subsequent launch while the access code is active.  The page then redirects to the `/index` page on the smart APP which will display the name of the patient in context.  It will use the code to query the keycloak token endpoint to get an actual access token which can be used to get resources from the fhir server.

It will then query the fhir server with the access token, which will introspect it and return the requested patient on success.

### DIFF
--- a/ehr-server/src/main/java/org/hl7/davinci/ehrserver/ClientAuthorizationInterceptor.java
+++ b/ehr-server/src/main/java/org/hl7/davinci/ehrserver/ClientAuthorizationInterceptor.java
@@ -28,8 +28,6 @@ public class ClientAuthorizationInterceptor extends AuthorizationInterceptor {
 
   @Override
   public List<IAuthRule> buildRuleList(RequestDetails theRequestDetails) {
-
-    System.out.println("henlo");
     String useOauth = Config.get("use_oauth");
     if (!Boolean.parseBoolean(useOauth)) {
       return new RuleBuilder()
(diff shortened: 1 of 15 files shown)


ehr-server/src/main/java/org/hl7/davinci/ehrserver/ClientAuthorizationInterceptor.java
```diff
@@ -28,23 +28,27 @@
 
   @Override
   public List<IAuthRule> buildRuleList(RequestDetails theRequestDetails) {
+
+    System.out.println("henlo");
```
eedrummer:
```suggestion

```

ehr-server/src/main/resources/fhirServer.properties
```diff
@@ -1,4 +1,6 @@
 client_id = app-token
-client_secret= #replaceMe#
+client_secret= af3057e5-f29e-4c15-83bc-63b4d0506df2
```
eedrummer:
```suggestion
client_secret= #replaceMe#
```

PR #75, opened 2018-11-19
Remove references to localhost in server admin UI

eedrummer:
AJAX requests are now to relative URLs and no longer contain a hostname or port.


KeeyanGhoreshi:
The fix I mentioned above turned out to be non-trivial.  So I went ahead and replaced the request to the `/api/public/` endpoint with a direct call to the public key repository object, since that is how it should have been implemented in the first place.

lpdecelles:
Looks good. Maybe try to improve logging in catch block.

PR #76, opened 2018-11-20
Updating the diagrams to reflect changes in project structure

eedrummer:
Renamed fhir-sever to ehr-server
Reflected changes in request generator moving into its own project
Added boxes to show payer and provider


KeeyanGhoreshi:
Is the SMART app intentionally left out of the diagram?

eedrummer:
For now, yes. I was updating this for the ONC Connectathon, where we will not cover DTR.

PR #78, opened 2018-11-29
fix smart url builder for patientid

daco101:
The PR tweaks the smart url builder to return a patient id and fhir server url that are ready to use by the js smart client.

### DIFF
--- a/server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/CdsService.java
+++ b/server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/CdsService.java
@@ -216,10 +216,10 @@ public abstract class CdsService<requestTypeT extends CdsRequest<?, ?>> {
 
   protected String smartUrlBuilder(String patientId, String fhirBase) {
     String launchUrl = myConfig.getLaunchUrl();
-    if (fhirBase.endsWith("/")) {
+    if (fhirBase != null && fhirBase.endsWith("/")) {
       fhirBase = fhirBase.substring(0, fhirBase.length() - 1);
     }
-    if (patientId.startsWith("Patient/")) {
+    if (patientId != null && patientId.startsWith("Patient/")) {
       patientId = patientId.substring(8,patientId.length());
     }
     return launchUrl + "?iss=" + fhirBase + "&patientId=" + patientId;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/CdsService.java
```diff
@@ -216,10 +216,15 @@ public CdsResponse handleRequest(@Valid @RequestBody requestTypeT request) {
 
   protected String smartUrlBuilder(String patientId, String fhirBase) {
     String launchUrl = myConfig.getLaunchUrl();
+    if (fhirBase != null && fhirBase.endsWith("/")) {
+      fhirBase = fhirBase.substring(0, fhirBase.length() - 1);
+    }
+    if (patientId != null && patientId.startsWith("Patient/")) {
+      patientId = patientId.substring(8,patientId.length());
```
KeeyanGhoreshi: I'm not sure if this is the right place to enforce this.  The server doesn't care what the reference is, our DTR app does.  Suppose someone implements their own SMART app to use with this server and they expect the reference to return as-is, like it's written in the FHIR.  I would argue that this should take place in the DTR app as well, but could be swayed by a strong counterargument.

KeeyanGhoreshi: But, counterpoint, if someone writes a SMART app where they expect the reference to return as-is, they would still know its a patient resource and could just as easily add on the `Patient/` prefix as we chop it off here.  Overall, I'm on the fence about what the best practice would be regarding where to do this chopping.

daco101: I would say that since both the js smartclient and the smarthealthit sandbox use/expect patient id without the "Patient/" prefix, that format is more likely to be the convention. I'm happy using the full "Patient/patientid" format, but then I think it is better to call it something like PatientResourceId.

KeeyanGhoreshi: if the js smartclient and smarthealthit sandbox use that format then I'd say we should do it like that as well.

PR #79, opened 2018-11-30
added build command to update node_interfaces

KeeyanGhoreshi:
Added a `buildx` command to the `node_interfaces` subproject which automatically moves the built out files into the correct folder so that they are used by the server when it is run.  Running `npm run buildx` or `npm run-script buildx` will build out `node_interfaces`, delete the old static files in the server's `static` folder, rename the built out .css and .js files, and move them to the proper directory.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -50,16 +50,6 @@ This functionality can be turned off and on by changing the `checkJWt` property 
 
 ## Building Web Apps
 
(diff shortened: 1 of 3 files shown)


lpdecelles:
Looks good, maybe we can add this to a post build event in the future under another Jira ticket.

PR #80, opened 2018-11-30
Replace code system url in the server UI - dmeerx-156

lpdecelles:


server/src/main/resources/node_interfaces/src/components/requestLog/HealthInfo.js
```diff
@@ -2,10 +2,8 @@ import React, {Component} from 'react';
 import './detail.css';
 import './request.css';
 
-const codeSystemConversion = {
-    "http://www.ama-assn.org/go/cpt":"CPT",
-    "https://bluebutton.cms.gov/resources/codesystem/hcpcs":"HCPCS"
-}
+let componentConstants = require('../ComponentConstants');
```
KeeyanGhoreshi: this works fine but could be streamlined as `import {codeSystemConversion} from './ComponentConstants'`.  Then, later in the code, you can do `codeSystemConversion[dataProps.codeSystem]` instead of `componentConstants.codeSystemConversion[dataProps.codeSystem]`

lpdecelles: Thanks Keeyan, I have made your suggestion and committed it. Larry

PR #83, opened 2018-12-07
removed unnecessary files

KeeyanGhoreshi:
Merge of the server ui changes from the huge PR had some extra files hiding in it that I forgot to delete. 

Also fixes a small bug with the EHR server where it will throw a null pointer if the client secret is wrong because it searched for a non-existent JSON key.


lpdecelles:
Keeyan,

This pull request has a couple conflicts. Can you please resolve these?

Thanks,
Larry

lpdecelles:
Keeyan,

The code looks good. How would you recommend I test this?

Larry

KeeyanGhoreshi:
There's nopt a pressing need to test it since it just deletes files that I accidentally committed to the repository.  If all the front end apps function correctly and the ehr-server correctly rejects a token when the client secret in fhirServer.properties is incorrect than everything should be good to go.

PR #88, opened 2018-12-21
New class to handle retrieving CQL rules from CDS Connect.

plarocque4:


lpdecelles:
Pat,

I am seeing errors when building (see screen shot below). Is there something missing in the project?

Note: I was able to build Keeyan's "cleanup" branch so think my environment is OK. 

Larry

screen shot 2018-12-31 at 2 25 18 pm

PR #89, opened 2018-12-28
[DME146] Openshift

KeeyanGhoreshi:
This PR adds support for openshift.  Most of the commits are small changes to the code that are quickly overwritten and changed.  The basic idea is that the projects relevant to the server (server + operations + resources) have a new `pom.xml` file along with an "aggergator" `pom.xml` in the parent directory that identifies the modules needed to run the server.  

The server `pom.xml` contains most of the changes.  It produces a single jar file when it builds the project which is what openshift uses to run the server.  Openshift was looking for jars, not wars, so this is why that change happened. Gradle will still output a war. 

All the `pom.xml` files are generated using gradle's maven plugin.  `gradle install` will create the pom.  For future consideration, it might be useful to have a script that automatically updates the `pom.xml` files when the gradle build changes, since the built out `pom.xml` is not sufficient and needs some static extras to go along with the dependencies to build the jar correctly.

There's an odd change in the commits where a line in `RequestLog` had to be deleted because it had non-ascii characters.  Someone wrote a comment with special quotation marks.

Also this thing runs on port 8080 now since that's what openshift wants to run it on.  I will probably have to change the port `ehr-server` runs on when I add it to openshift.  

Some notes on the openshift build for future readme editing (after the ehr-server & smart app are hopefully run in openshift as well). The app is built with red hat openjdk 8 image.  It uses the base directory as its context directory and pulls from this branch (the openshift branch).  After this gets merged in I'll have to change it to master.  The build needs to be configured such that the environment variable `ARTIFACT_DIR` is set to `server/target`.  This is what lets all three modules get built but only the `server` gets deployed.


KeeyanGhoreshi:
Openshift-ehr branch merge adds the pom for ehr-server that openshift uses to run the ehr server. Besides some small fixes to the pom to get it to be compliant, the only thing that really needed to be done was specify the `finalName` in the maven build so that the path to the server would be ehr-server instead of what it apparently used to be - /ehr-server-unspecified/r4.

Some notes for setting it up in openshift:

Tomcat 8 builds but eventually crashes openshift because it makes, and then logs, some proxy call every 10 seconds forever.
JBoss Web Server 3.1 Tomcat 7 runs it successfully with no weird logging bugs.

server/src/main/resources/application.yml
```diff
@@ -13,7 +13,7 @@ spring:
       hibernate.dialect: org.hibernate.dialect.H2Dialect
       hibernate.hbm2ddl.auto: create-drop
 server:
-  port: 8090
+  port: 8080
```
eedrummer: This is likely to break things, like the `crd-request-generator`. I think we either need a strategy for running both the ehr-server and this on the same machine or figure out a way to make OpenShift happy with it running on `8090`.

KeeyanGhoreshi: I got openshift to run it on 8090 so we should be good now.  For reference, you gotta change the service and deployment YAMLs to use port 8090.

lpdecelles:
Keeyan, It looks like you made the changes Andy suggested. Is that correct?

PR #91, opened 2019-01-03
Openshift ehr

KeeyanGhoreshi:
This one adds the pom for `ehr-server` that openshift uses to run the ehr server.  Besides some small fixes to the pom to get it to be compliant, the only thing that really needed to be done was specify the `finalName` in the maven build so that the path to the server would be `ehr-server` instead of what it apparently used to be - `/ehr-server-unspecified/r4`.  

Some notes for setting it up in openshift:

- Tomcat 8 builds but eventually crashes openshift because it makes, and then logs, some proxy call every 10 seconds forever.
- JBoss Web Server 3.1 Tomcat 7 runs it successfully with no weird logging bugs.


KeeyanGhoreshi:
Looks good to me

PR #95, opened 2019-01-09
small changes

KeeyanGhoreshi:


KeeyanGhoreshi:
We should probably find some way for the react to update and build out automatically.

PR #96, opened 2019-01-14
[DME 195,165,152] Adding cql rules

daco101:
This PR adds support for CQL rules and their execution.
Todo after this: update front end, update logger, figure out how we are dealing with cql libraries and implement the library loading associated with it, allow us to get cql rules from cds connect and switch between local db and cds connect for rules.


server/src/main/resources/rules/cms/hcpcs/E0250.cql
```diff
@@ -0,0 +1,34 @@
...
+  AgeInYears()
+
+define "Gender":
+  Patient.gender.value
+
+define RULE_APPLIES:
+  true
+
+define "Auth_Needed":
+  "Gender" = 'male'
```
plarocque4: Is the Gender used in Auth_Needed or RULE_APPLIES?

lpdecelles:
Hey Cole, It looks like this PR has conflicts. Can you please resolve these? Thanks, Larry

PR #103, opened 2019-01-28
[202] Update fhir library

KeeyanGhoreshi:
Updates the HAPI FHIR library we use from 3.4 to 3.6.  The only notable change is that `ElegibilityRequest` was apparently renamed to `CoverageElegibilityRequest`, same with `ElegibilityResponse`.  

Also made some changes to the server gradle file to get it to build since we were getting multiple versions of the same libraries which caused issues.  Not sure how we avoided this problem till now, but I had to remove a couple libraries that spring boot was already providing and upgrade the version of spring boot we're using.


lpdecelles:
Keeyan, it looks like this has conflicts.

lpdecelles:
Hey Keeyan, It looks like this branch has conflicts. Can you please fix this?

PR #104, opened 2019-01-30
[201] Price Details

daco101:
This PR adds a price details column (and the dummy data) and prepends that to the details of returned cards.


server/src/test/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/MedicationPrescribeServiceTest.java
```diff
@@ -31,6 +32,8 @@ public void testHandleRequest() {
     CdsResponse response = service.handleRequest(request);
     assertNotNull(response);
     assertEquals(1, response.getCards().size());
-    assertEquals("Documentation is required for the desired device or service", response.getCards().get(0).getSummary());
+    assertEquals("Documentation is required for the desired device or service.", response.getCards().get(0).getSummary());
+
+    assertTrue(response.getCards().get(0).getDetail().startsWith("Price details: Payor will cover up to $680."));
```
KeeyanGhoreshi: This test will have to be rewritten if the rule changes.  It might be worth it to either search the data sql for the correct payment text or have the request be for a fake entry that gets added to the database by this test instead of relying on stuff that's "supposed" to be in the database.

server/src/main/java/org/hl7/davinci/endpoint/components/CardBuilder.java
```diff
@@ -51,7 +52,8 @@ public static Card transform(CoverageRequirementRule crr, String launchUrl) {
       }
       String patientStateRule = Optional.ofNullable(crr.getPatientAddressState()).orElse("Any");
       String providerStateRule = Optional.ofNullable(crr.getProviderAddressState()).orElse("Any");
-      card.setDetail(String.format(detail, genderRule, crr.getAgeRangeLow(),
+      card.setDetail("Price details: " + crr.getPriceDescription() + "\n " +
```
KeeyanGhoreshi: In Github markdown that `\n` will not get shown as an actual line break.  You'll need to do either `__\n` (the underscores represent spaces, you need two spaces to get a new line) or an explicit `<br />`.

daco101:
Are you saying we need to create new rules for all of the items in the Limited_Pricing file? Is this for himss? Should the rules match anything in particular?

KeeyanGhoreshi:
When you get home oxygen it should display all related codes, the table isn't a list of codes and their prices it's a list of related codes that are all displayed together to show the code you're asking for and its accessories.  So if we were implementing this for real we'd have a database with all the codes and their relation to one another and when you order something the details section would have a table with ALL the related codes and their prices to give an idea of the total price (or other options).

daco101:
I see so the rule for E0424 should have the entire table from that file, but for all the other rules we have no pricing info yet.

PR #105, opened 2019-01-30
[174] Scenario integration tests

KeeyanGhoreshi:
Adds an integration test based on the connectathon scenarios. 

Basically just sends a request for every possible combo of prefetch/noprefetch for a rule that requires or does not require documentation for both medication and device request (8 combos!).  

Also makes a gradle task to run the integration test which starts the EHR server, runs the test, and then shuts the EHR server down.  

The integration test will only run if you run the "integrationTest" gradle task in the server project.


lpdecelles:
Should we add integrationTest stuff to the readme?

KeeyanGhoreshi:
Recent changes to master break this branch.  I'll make a new PR for the HIMSS master.  Also, the CQL switch is breaking a ton of stuff in master, it should be in a separate branch.

PR #106, opened 2019-01-30
[213] Request log pages

KeeyanGhoreshi:
Fixes the little page number buttons on the request log.  It now only shows 9 pages at a time, and there are little arrow buttons that you can use to scroll through the pages.

Also fixed inconsistent button sizing and centering.

Testing might be a bit difficult since you need to run the server version.  I haven't included the static files in this PR since the structure of the static files gets changed in my other PR #102 and they will have to be rebuilt anyway (and also my react scripts was installed globally and now builds it the new way so I can't get the old static scripts to even upload here)

If you want to test it, run `buildout.sh` first and then run the server.  You'll have to make over 150 requests to see the new changes in action, or you can change `entriesPerPage` at the top of `RequestLog.js` to reduce the amount of requests it displays per page and then build that out into static files with `buildout.sh`. Then you can get away with 10 requests at a `entriesPerPage` of 1.


lpdecelles:
Should we put this in the readme?

somemachine:node_interfaces someuser$ sh buildout.sh

PR #109, opened 2019-02-17
[223] Zip Rules

daco101:
Moved rules to zip files, added support for extracting cql files from them.
Added support for packaging libraries with cql libraries.


plarocque4:
Looks good

PR #114, opened 2019-03-08
Updates to CDS Connect interface after end to end testing.

plarocque4:


lpdecelles:
Hello Pat,

I looks like Travis CI is failing.

Larry

server/src/main/java/org/hl7/davinci/endpoint/cql/bundle/CqlBundle.java
```diff
@@ -64,6 +69,24 @@ public RawCqlLibrarySourceProvider getRawCqlLibrarySourceProvider() {
...
+  public static boolean checkIfZip(File file) {
+    try {
+      RandomAccessFile f = new RandomAccessFile(file, "r");
+      long n = f.readInt();
+      f.close();
+      // check the header to see if it is a zip file
+      if (n == 0x504B0304) {
+        return true;
+      }
+    } catch (FileNotFoundException e) {
```
eedrummer: I realize that if a file is not found, it's going to result in a RuntimeException further down the stack. That said, throwing the FNFE here might help with debugging later if this code ever becomes an issue. You could have the code claiming that the bundle is not a zip file, when it really just can't find the file.

server/src/main/java/org/hl7/davinci/endpoint/cdsconnect/CdsConnectRuleList.java
```diff
@@ -35,4 +44,43 @@ public CdsConnectRuleList(CdsConnectConnection connection, String data) {
...
+          CoverageRequirementRule rule = new CoverageRequirementRule();
+          rule.setId(nodeId).setPayor(payor).setCodeSystem(codeSystem).setCode(code);
+
+          rule.setCqlPackagePath("unknown");
+          CqlBundle emptyCqlBundle = new CqlBundle();
+          emptyCqlBundle.setRawMainCqlLibrary("unknown");
+          rule.setCqlBundle(emptyCqlBundle);
+          rules.add(rule);
+
+        } catch (NullPointerException e) {
```
eedrummer: My guess is that this is to handle cases where the JSON doesn't contain properties that you expect. I'd recommend using the `has` method to check instead of catching the exception. You could trade the try / catch block for an if statement looking for the three properties you are pulling from the JSON. That way, you won't accidentally swallow an NPE on code you didn't expect to throw it.

PR #120, opened 2019-03-27
[239] Add a link to the rules UI for linking the user to the relevant…

plarocque4:
… CDS Connect Artifact


server/src/main/java/org/hl7/davinci/endpoint/cdsconnect/CdsConnectConnection.java
```diff
@@ -213,4 +213,46 @@ public String retrieveArtifact(Integer nodeId) {
 
     return fileResponse.getBody();
   }
+
+  public String getRuleLink(String title) {
+    String titleFixed = title.replace(' ','-')
```
KeeyanGhoreshi: Could this be some kind of regex?  `title.replaceAll("[^A-Za-z0-9]", "");`?

server/src/main/java/org/hl7/davinci/endpoint/cdsconnect/CdsConnectConnection.java
```diff
@@ -213,4 +213,46 @@ public String retrieveArtifact(Integer nodeId) {
...
+        .replace(")", "")
+        .replace("+", "")
+        .replace("=", "")
+        .replace("_", "")
+        .replace("--", "-")
+        .toLowerCase();
+
+    // strip the last '-' off the string if it is the final character
+    while (titleFixed.endsWith("-")) {
+      titleFixed = titleFixed.substring(0, titleFixed.length()-1);
```
KeeyanGhoreshi: Maybe I'm misunderstanding, but wouldn't a `-` only be the final character if it replaced white space in the statement above?  Wouldn't it be better to `trim()` the thing first?  Or is there cases where there are lots of hyphens at the end of the title?  You could also replace `-` with white space, `trim()`, and then replace whitespace, like so: `a.replace("-"," ").trim().replace(" ","-");`.

server/src/main/java/org/hl7/davinci/endpoint/cdsconnect/CdsConnectRuleList.java
```diff
@@ -85,12 +85,20 @@ public CdsConnectRuleList(CdsConnectConnection connection, String data) {
           continue;
         }
 
+        if (jsonRuleObject.has("title")) {
+          String title = jsonRuleObject.get("title").getAsString();
+          String editLink = connection.getRuleLink(title);
+          rule.setEditLink(editLink);
+        } else {
+          logger.info("could not find titel, skpping rule with node id " + nodeId);
```
KeeyanGhoreshi: titel should be title I'm assuming

PR #133, opened 2019-05-21
Change launch

KeeyanGhoreshi:
The large majority of these commits are changes to the bundled js files, which were being committed to push up to heroku to change the embedded DTR server.  Otherwise there are changes to accommodate Epic EHR, like changing the app context to a string and passing the device request through the app context.


server/src/main/jib/rules/cms/hcpcs/E0424/E0424.cql
```diff
@@ -20,9 +20,7 @@ define "ProviderState":
 
 
 define RULE_APPLIES:
-  "Gender" = 'male'
-  and "PatientHomeState" = 'TX'
-  and "ProviderState" = 'TX'
+  "Age" > 0
```
plarocque4: Is there any reasoning around what we want to have in these rules? So far they seem to be whatever we have felt like putting in to make it easier to test.

PR #139, opened 2019-06-12
Add in Positive Airway Pressure device

mulcahyk:


server/src/main/java/org/hl7/davinci/endpoint/GeneratePapQuestionnaire.java
```diff
@@ -0,0 +1,264 @@
+package org.hl7.davinci.endpoint;
```
plarocque4: I don't think that this file belongs in endpoint since it is a standalone application and not an endpoint. Maybe a utilities or builder folder would be a better place for it.

plarocque4: Looks good otherwise.

PR #146, opened 2019-10-31
[353] Add FHIR R4 support to CRD

plarocque4:


lpdecelles:
image

Pat, it seems to work. But how is it working without an R4 ehr-server?

plarocque4:
> image
> 
> Pat, it seems to work. But how is it working without an R4 ehr-server?

You selected "Include Prefetch" so it should not have needed to reach back to the EHR-Server for the resources. The request generator already has r4 resources hard coded in it. This will hopefully change with @KeeyanGhoreshi changes that will make the request generator pull patients from the EHR Server.

PR #147, opened 2019-11-05
Initial pass at refactoring DTR CQL. Multiple libraries and value sets

hossenlopp:
- Moved basic patient info out of the two existing pre-population libraries
- Created shared BasicPatientInfo library
- Made both questionnaires reference the basic patient info library
- Updated CQL result references in questionnaires to reference library name and statement name
- Started ValueSet refactor by moving COPD codes in OxygenTherapy to an "expanded" FHIR ValueSet.
- Added dataRequirement to the oxygen therapy library to reference the ValueSet


lpdecelles:
It works! Need to review the code still.

lpdecelles:
OK, I have created DMEERX-406 to address Keeyan's request above.

PR #148, opened 2019-11-08
Update FHIR Resources with missing data. 

plarocque4:
Update CardBuilder::setRuleApplies to handle null values.


lpdecelles:
This seems to fix things, so I am merging it.

PR #151, opened 2019-11-19
Downgrade cql-to-elm and resolve CQL warnings

hossenlopp:


lpdecelles:
This is no longer needed

PR #161, opened 2019-12-12
Papcrdrules

yuliumitre:
https://jira.mitre.org/browse/DMEERX-377
Convert CQL rules in CRD to real PAP rules - maintain STU 3 support

1. Add CRD rules for E0470 and E0601 R4 and STU3


server/src/main/jib/rules/cms/hcpcs/E0470/E0470-r4.cql
```diff
@@ -0,0 +1,27 @@
...
+include FHIRHelpers version '4.0.0' called FHIRHelpers
+
+parameter Patient Patient
+parameter device_request DeviceRequest
+
+define "Age":
+AgeInYears()
+
+define RULE_APPLIES:
+    "Age" > 0
```
plarocque4: Does this work if you just return true?

server/src/main/jib/rules/cms/hcpcs/E0470/E0470-r4.cql
```diff
@@ -0,0 +1,27 @@
...
+
+define RESULT_Summary:
+  'Authorization not required.'
```
plarocque4: Perhaps these should just say "Documentation Required"

server/src/main/jib/rules/cms/hcpcs/E0470/E0470-r4.cql
```diff
@@ -0,0 +1,27 @@
...
+
+define RESULT_Details:
+  'Authorization is not required.'
```
plarocque4: Maybe this could say: "Documentation Required, please complete form via 'Smart App' link"

PR #166, opened 2019-12-31
Paddtrrules bp

yuliumitre:
1. Add R4 DTR rules for PAP
2. Using BasicPractitionerInfo library


server/src/main/jib/smartAppFhirArtifacts/positive-airway-pressure-questionnaire-r4.json
```diff
@@ -0,0 +1,423 @@
...
+        "reference": "urn:hl7:davinci:crd:library-positive-airway-pressure-prepopulate-r4"
+      }
+    },
+    {
+      "url": "http://hl7.org/fhir/StructureDefinition/cqif-library",
+      "valueReference": {
+        "reference": "urn:hl7:davinci:crd:library-basic-patient-info-prepopulate-r4"
+      }
+    }
+  ],
```
hossenlopp: This is missing a reference to the `urn:hl7:davinci:crd:library-basic-practitioner-info-prepopulate` library.

yuliumitre: Good catch 👍  This should be fixed now.

PR #178, opened 2020-02-06
[410] Display if Prior Auth or Documentation Requirements only is required

plarocque4:
Update rules to include flag for prior auth and documentation. Pass the prior auth flag to DTR and suppress smart app link if neither are required.


lpdecelles:
Seems to work. Still need to tweak text (Prior Auth needed, etc.)

PR #179, opened 2020-02-10
Hbcrdrules

yuliumitre:
Update CRD rules for hospital beds order.


lpdecelles:
image

Yolanda we think this is failing because of a "text string" in the CI test. You can run these with this: "gradle clean check" which is part of the readme see the failure.

yuliumitre:
I will take a look at the issues.

lpdecelles:
It might be that we need to change the test(s). Maybe to not be as strict.

yuliumitre:
@lpdecelles The test has been fixed.

lpdecelles:
image

This looks good a card is returned.

PR #183, opened 2020-02-14
DMEERX-427 Ventilator Order rules

yunwwang:


lpdecelles:
image

I was not able to select E0465 from the request generator in order to test.

lpdecelles:
image

I was able to use Eo465, but card is returned. See image above.

lpdecelles:
Looks good. I unit tested with E0465. Also need to update the readme to show CRD starts with "gradle server:bootRun"

PR #184, opened 2020-02-25
Hbdtrrules

yuliumitre:
Add hospital bed dtr rules.

To test this PR, need to switch to the same name hbdtrrules branch of test-ehr and dtr repo. 

Every patient has the device request for hospital bed. Patient pat014 has two encounters.


lpdecelles:
Looks good. Still have a few missing pre-populated fields.

PR #185, opened 2020-02-25
DMEERX-427 ventilator

yunwwang:
Update CQL rules for dynamic options 
Add itemControl to Questionnaire


lpdecelles:
Looks good. I know there are some fields that still need to be pre-populated. Like the new table etc.

PR #186, opened 2020-02-28
For HIMSS prepopulate more fields in oxygen therapy using cql

hossenlopp:


server/src/main/jib/smartAppFhirArtifacts/OxygenTherapyPrepopulation.cql
```diff
@@ -149,7 +179,19 @@ define PatientMobile: not exists(Confirmed(ActiveOrRecurring([Condition: "Immobi
 define HematocritThatIsGreaterThanThreshold: HighestObservation(WithUnit(Verified(ObservationLookBack([Observation: "Hematocrit_lab_test_Codes"], 3 months)), '%'))
 define PatientHasHematocritThatIsGreaterThanThreshold: exists("HematocritThatIsGreaterThanThreshold")
 
+// prescribed use
+define OccurrenceTimingDuration: (device_request.occurrence as FHIR.Timing).repeat.bounds as FHIR.Duration
```
yuliumitre: In general, double quoting the define name to be consistent.

server/src/main/jib/smartAppFhirArtifacts/OxygenTherapyPrepopulation.cql
```diff
@@ -149,7 +179,19 @@ define PatientMobile: not exists(Confirmed(ActiveOrRecurring([Condition: "Immobi
...
+define OccurrenceTimingQuantity:
+  if OccurrenceTimingDuration is not null
+  then System.Quantity { value: OccurrenceTimingDuration.value.value, unit: OccurrenceTimingDuration.unit.value }
+  else null
+define LengthOfNeed: 
+  ConvertQuantity(OccurrenceTimingQuantity, 'mo').value
 
+define FrequencyOfUse:
+  if (device_request.occurrence as FHIR.Timing).repeat."when" = 'HS'
+    then { 'During sleep' }
```
yuliumitre: This is more of a question. Can we take advantage of the display of the valueset? As seems like we have to explicitly know the display and hard code here.

hossenlopp: The hard code is to match what the questionnaire is expecting. The `code` is just a FHIR code which is just a string there is no display info.

yuliumitre: okay, got it. Thanks.

hossenlopp: I decided to put the code in the occurenceTiming.code in the data, just incase we wanted to use that. But the the occurenceTiming.repeat.when seemed more appropriate for now. So the CQL is setup to use that for now.

I just noticed the cardinality of Timing.repeat.when is 0..*. I should update the logic to look at all of these codes and return the questionnaire expected responses for any we can decipher.

server/src/main/jib/smartAppFhirArtifacts/home-oxygen-questionnaire_2.json
```diff
@@ -548,6 +560,7 @@
           "text": "Type",
           "type": "choice",
           "required": true,
+          "readOnly": true,
```
KeeyanGhoreshi: Why was this made read only?

hossenlopp: Because it is now determined by the CQL just like the stationary vs portable question.

server/src/main/jib/smartAppFhirArtifacts/OxygenTherapyPrepopulation.cql
```diff
@@ -128,73 +158,99 @@ define "Immobilization_Codes": { "102491009", "129857008", "129859006", "1606850
...
-define DeviceRequestHcpcsCoding: singleton from (
+define "DeviceRequestHcpcsCoding": singleton from (
   ((cast device_request.code as CodeableConcept).coding) coding
     where coding.system.value = 'https://bluebutton.cms.gov/resources/codesystem/hcpcs')
 
-define DeviceRequestDescription: 'HCPCS ' + "DeviceRequestHcpcsCoding".code.value + ' - ' + "DeviceRequestHcpcsCoding".display.value
-define DeviceRequestedIsPortable: "DeviceRequestHcpcsCoding".code.value in { 'E0433', 'E0434', 'E0444', 'EO431', 'K0738', 'E0443', 'E1392' }
-define DeviceRequestedIsStationary: "DeviceRequestHcpcsCoding".code.value in { 'E0439', 'E0442', 'E0424', 'E0441', 'E1390', 'E1391' }
-
+define "DeviceRequestDescription": 'HCPCS ' + "DeviceRequestHcpcsCoding".code.value + ' - ' + "DeviceRequestHcpcsCoding".display.value
```
KeeyanGhoreshi: The display value isn't always included and when it isn't there it causes the field not to be filled, for example with the Rush device request the display name is not included.  Is there some kind of null-handling we can do to avoid that?

hossenlopp: Check the latest commit. It matches the code with the copy of the code in the CQL and will use it's display text if the DeviceRequest doesn't have the display.

PR #187, opened 2020-02-28
[496, 497, 498, 504, 648] First set of work supporting new file storage.

plarocque4:
The rules files are now located in a separate repo: https://github.com/HL7-DaVinci/CDS-Library
The localDb file and GitHub can be used for CRD only, they are not yet capable of grabbing files necessary for DTR. The file retrieval for DTR still uses the old local file storage location from before.
This is a breaking change that does not yet support CDS Connect in any way.
Files for DTR are only retrieved from the local folder within the jib/smartAppFhirArtifacts folder.
There is also a new REST endpoint <server>/reload that forces the lookup table to be rebuilt.


server/build.gradle
```diff
@@ -104,3 +104,31 @@ void embedDtrScript() {
...
+    }
+}
+void embedCdsLibraryScript() {
+    exec {
+        workingDir './'
+        commandLine 'git', 'clone', 'https://github.com/HL7-DaVinci/CDS-Library.git', 'CDS-Library'
+    }
+    exec {
+        workingDir './CDS-Library'
+        commandLine 'git', 'checkout', 'master'
```
yuliumitre: We should allow to point to different branch. When we develop rulesets, we need to use different branch from master. 

Also, we need to allow override the existing CDS-Library directory if it already exists.

plarocque4: Are you looking for a parameter to the script so you could provide the branch? If a different branch was desired, you could always modify the script or simply pull it down somewhere yourself and point the application.yml at it.

yuliumitre: Yeah, a parameter would be helpful :-)

server/src/main/java/org/hl7/davinci/endpoint/files/local/LocalFileStore.java
```diff
@@ -0,0 +1,210 @@
...
+    String localPath = config.getLocalDb().getPath();
+    String cqlFileLocation = localPath + topic + "/" + fhirVersion + "/files/";
+    File dir = new File(cqlFileLocation);
+    String regex = name + "-\\d.\\d.\\d" + extension;
+    FileFilter fileFilter = new RegexFileFilter(regex);
+    File[] files = dir.listFiles(fileFilter);
+    for (int i=0; i < files.length; i++) {
+      //logger.info("LocalFileStore::findFile(): matched file: " + files[i]);
+      return files[i];
+    }
```
yuliumitre: Do we just assume there is only one file? Can we do files[0]?

server/src/main/java/org/hl7/davinci/endpoint/files/local/LocalFileStore.java
```diff
@@ -0,0 +1,210 @@
...
+                // convert to object
+                TopicMetadata metadata = objectMapper.readValue(content, TopicMetadata.class);
+
+                for (Mapping mapping: metadata.getMappings()) {
+                  for (String code: mapping.getCodes()) {
+                    for (String payer: metadata.getPayers()) {
+                      for (String fhirVersion: metadata.getFhirVersions()) {
+
+                        String mainCqlLibraryName = metadata.getTopic() + "Rule";
+                        File mainCqlFile = findFile(metadata.getTopic(), fhirVersion, mainCqlLibraryName, ".cql");
```
yuliumitre: Have a question about the rule file name? What does those version stand for? E.g. HospitalBedsRule-0.1.0.cql  Do we plan to have different versions of CQL file?

server/src/main/java/org/hl7/davinci/endpoint/rules/CoverageRequirementRuleCriteria.java
```diff
@@ -10,6 +10,7 @@
   private String payor;
   private String code;
   private String codeSystem;
+  private String fhirVersion;
```
yuliumitre: I think we are returning multiple cards for the same code under different topics. Another option is to add topic to the criteria for specific rule. Could be future improvement.

yuliumitre:
The PR looks pretty good. Minor comments and some questions. Thanks.

server/src/main/java/org/hl7/davinci/endpoint/rules/CoverageRequirementRuleCriteria.java
plarocque4: That would require us to know which topic we need from the information coming in the original CRD request. I'm not sure that is stored anywhere, just which device is desired.

server/src/main/java/org/hl7/davinci/endpoint/files/local/LocalFileStore.java
plarocque4: This came out of a discussion with @hossenlopp, we decided to name the files using the name and version that are found in the first line of the file. This does mean that the version number would be updated in the future. The code looks for the first file matching the right name and ignores the version number because it expects there to be only one. In the future it might be a good addition to match the name and find the file with the highest version number.

server/src/main/java/org/hl7/davinci/endpoint/files/local/LocalFileStore.java
plarocque4: Yes that's true, this used to log all the files it found that's why it was written this way. I'll make sure to update it.

server/src/main/java/org/hl7/davinci/endpoint/rules/CoverageRequirementRuleCriteria.java
yuliumitre: Okay. It is fair for this story. As the new RAD CRD comes in, we might need to make changes accordingly.

server/src/main/java/org/hl7/davinci/endpoint/files/local/LocalFileStore.java
yuliumitre: Sounds good.

lpdecelles:
image

I am seeing this error when trying to start CRD "gradle server:bootRun"

PR #188, opened 2020-03-02
Neatdtrrules.hk

hkong:
Updated NEAT CQL and questionaire


plarocque4:
The changes in this PR will now be moved to the CDS-Library repo where all of the rule set files (CQL & FHIR Resources) now belong.

PR #200, opened 2020-05-04
[729] Get new DRLS branch working with Logica Health - Connectathon

plarocque4:
Update inserted link to work through a proxied web server.

### DIFF
--- a/server/src/main/java/org/hl7/davinci/endpoint/Utils.java
+++ b/server/src/main/java/org/hl7/davinci/endpoint/Utils.java
@@ -9,21 +9,24 @@ public class Utils {
 
   public static URL getApplicationBaseUrl(HttpServletRequest request) {
     try {
-      // grab the forwarded values if they are not null
-      String serverName = ((request.getHeader(HttpHeaders.X_FORWARDED_HOST) != null) ? request.getHeader(HttpHeaders.X_FORWARDED_HOST) : request.getServerName());
-      String scheme = ((request.getHeader(HttpHeaders.X_FORWARDED_PROTO) != null) ? request.getHeader(HttpHeaders.X_FORWARDED_PROTO) : request.getScheme());
-      int port = ((request.getHeader(HttpHeaders.X_FORWARDED_PORT) != null) ? Integer.parseInt(request.getHeader(HttpHeaders.X_FORWARDED_PORT)) : request.getServerPort());
       URL url = null;
 
-      // grab the last forwarded url
-      String[] serverParts = serverName.split(", ");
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/hl7/davinci/endpoint/Utils.java
```diff
@@ -3,12 +3,29 @@
...
     try {
-      URL url = new URL(request.getScheme(), request.getServerName(), request.getServerPort(), request.getContextPath());
+      // grab the forwarded values if they are not null
+      String serverName = ((request.getHeader(HttpHeaders.X_FORWARDED_HOST) != null) ? request.getHeader(HttpHeaders.X_FORWARDED_HOST) : request.getServerName());
+      String scheme = ((request.getHeader(HttpHeaders.X_FORWARDED_PROTO) != null) ? request.getHeader(HttpHeaders.X_FORWARDED_PROTO) : request.getScheme());
+      int port = request.getServerPort();
+      URL url = null;
+
+      // grab the last forwarded url
+      String[] serverParts = serverName.split(", ");
```
hossenlopp, line 19: Do we know that there are multiple forwarded hosts? This should definitely work fine if there is just one.

server/src/main/java/org/hl7/davinci/endpoint/Utils.java
```diff
@@ -3,12 +3,29 @@
...
+      serverName = serverParts[serverParts.length - 1];
+
+      // set the port if not the default http port (80)
+      if (port != 80) {
```
hossenlopp, line 23: I think using the port coming from `request.getServerPort()` may be problematic. When there is no proxy in front of CRD, it's fine. But I think we may need to look for a port in `X-Forwarded-Host` or `X-Forwarded-Port` to find the real port that the request went to. request.getServerPort() is probably going to be the port the proxy server reached CRD at.

plarocque4: The port was set properly on logicahealth when I tried this before. I will look into the X-Forwarded-Port though.

hossenlopp: Ah. So we are running it at 80 and the logicahealth stuff is proxying https requests at standard 443 to us at 80?

plarocque4: Yes, that is correct.

plarocque4:
I have updated to use the X_FORWARDED_PORT if available but it is not set on LogicaHealth.

PR #201, opened 2020-05-07
DMEERX-739 - update capability statement for order-sign and order-select

lpdecelles:
Added order-sign and order-select to capability statement


plarocque4:
I think we need to update the capability statement for STU3 as well.

plarocque4:
I feel silly for missing this, thanks for fixing!

lpdecelles:
@plarocque4 did we do this STU 3 also?

plarocque4:
> @plarocque4 no worries, it not a big deal. Did we do this STU 3 also?

Yes, we support all four of them for STU3 as well.

lpdecelles:
I have made the same fix for STU 3

PR #202, opened 2020-05-18
714 - ValueSets in Questionnaires

hossenlopp:
Stores the url fields of ValueSets when loaded and adds a way to search by the urls. This adds support for valuesets from other sources with canonical urls to be used as long as they can be pulled in to our CDS-Library repo.

Embeds referenced valuesets in questionnaires and provides an abstract class for doing generic modifications on FHIR resources.

An example of using this work with the HL7 FHIR administrative-gender valueset in the HOT questionnaire can be seen by using the `hot-valuesets-questionnaire` branch of CDS-Library.

### DIFF
--- a/server/src/main/java/org/hl7/davinci/endpoint/files/QuestionnaireValueSetProcessor.java
+++ b/server/src/main/java/org/hl7/davinci/endpoint/files/QuestionnaireValueSetProcessor.java
@@ -53,19 +53,40 @@ public class QuestionnaireValueSetProcessor extends FhirResourceProcessor<Questi
     for (QuestionnaireItemComponent itemComponent : itemComponents) {
       // If there is an answerValueSet field we need to do some work on this item
       if (itemComponent.hasAnswerValueSet()) {
-        
+
+        // Only look for a valueset to embed if it does not appear to be a hash reference
         if (!itemComponent.getAnswerValueSet().startsWith("#")) {
           logger.info("answerValueSet found with url - " + itemComponent.getAnswerValueSet());
           String valueSetId = findAndLoadValueSet(itemComponent.getAnswerValueSet(), valueSetMap, fileStore, baseUrl);
-          itemComponent.setAnswerValueSet("#" + valueSetId);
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/hl7/davinci/endpoint/files/QuestionnaireValueSetProcessor.java
```diff
@@ -0,0 +1,110 @@
...
+          String valueSetId = findAndLoadValueSet(itemComponent.getAnswerValueSet(), valueSetMap, fileStore, baseUrl);
+          itemComponent.setAnswerValueSet("#" + valueSetId);
+          logger.info("answerValueSet replaced with  - " + itemComponent.getAnswerValueSet());
+        }
+
+        // Add initial reference to any item with an answerValueSet due to HAPI encoder error.
+        // This is needed to make sure the referenced ValueSets in contains are included.
+        QuestionnaireItemInitialComponent initial = new QuestionnaireItemInitialComponent(
+          new Reference(itemComponent.getAnswerValueSet()));
+        itemComponent.addInitial(initial);
```
yunwwang, line 68: If the item already have an initalValueSet reference to the contained valueset (not likely but possible), then skip adding another one

yunwwang: A Questionnaire may have pre existing contained value sets for some small non standard codings. For example, yes/no answer value sets. Will this handle (or skip) those?

hossenlopp: Line 57 skips the existing contained references, but still adds the initialValueSet. I didn't think about the chance that an initial value already exists... I'll think about that more... I wonder if there is another place that we can reference the ValueSets to make sure they are included but doesn't affect the initial values at all.

yunwwang: My use cases is limited to the current item. The goal to to prevent duplicated initial values in current item. I don't think global search is necessary.

PR #203, opened 2020-05-21
689 Support of sub-questionnaire

yunwwang:
Add methods to assemble from sub-questionnaires


lpdecelles:
Looks good. If we add more sub-questionnaries later, do we need to change CRD?

yuliumitre:
The changes look good to me. 👍

yunwwang:
These changes covers those simple replacement. It does not cover nested sub-questionnaire nor multiple root items in sub questionnaire. We can update the logic when we need to process such sub questionnaire in the future.

PR #206, opened 2020-06-03
Fix Extension comparasion

yunwwang:
This PR fixes some bugs in sub-questionnaire processing after cqf library changes:
1) Added capability to import more than one root items from sub questionnaire.
2) Fix bugs in comparing Extensions related to cqf canonical url

### DIFF
--- a/server/src/main/resources/application.yml
+++ b/server/src/main/resources/application.yml
@@ -55,7 +55,7 @@ gitHubConfig:
   useZipForReload: true
 
 localDb:
-  path: ../../CDS-Library/
+  path: CDS-Library/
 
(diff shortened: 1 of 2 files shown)


server/src/main/resources/application.yml
```diff
@@ -55,7 +55,7 @@ gitHubConfig:
   useZipForReload: true
 
 localDb:
-  path: CDS-Library/
+  path: ../../CDS-Library/
```
hossenlopp, line 58: This change shouldn't be made.

yunwwang: Good catch.!

PR #222, opened 2020-11-16
remove extra code

KeeyanGhoreshi:
Originally thought we needed some bigger changes to our JWT filter due to some errors I was seeing but it turns out I just had the Spring magic wrong.


server/src/main/resources/application.yml
```diff
@@ -23,7 +23,7 @@ server:
     # contextPath should start with "/" but not end with one. This is good: "/cds-hooks"
     contextPath: /
 
-checkJwt: false
+checkJwt: true
```
plarocque4, line 26: Do we want to change the default here to always check for JWT?

PR #223, opened 2020-12-09
Changes to work with LogicaHealth Sandbox

plarocque4:
Remove the UUID (36 character) requirement on the hookInstance.
Move the user from the hook to the userId in the context of the hook where it should be.


KeeyanGhoreshi (commented):
code changes look good, how can I test this?

plarocque4:
> code changes look good, how can I test this?

Just submit a CRD request from the request generator. Will need to do this with the matching branch on the request generator. The branch is dockerize_userid

KeeyanGhoreshi (approved):

PR #224, opened 2020-12-10
Update VSAC / UMLS login credentials to use API Key

plarocque4:
The Username / Password method has been deprecated. The implementation now uses the API Key that is found at the UMLS account site.


KeeyanGhoreshi (approved):
seems pretty straightforward to me

PR #228, opened 2021-02-19
DMEERX-794 Validate Mettles Rule Sets

yuliumitre:
This PR modifies CRD for make Mettles rule sets work. 

To test, use test-ehr and CDS-Libary prosrule branches.

### DIFF
--- a/resources/src/main/java/org/hl7/ShortNameMaps.java
+++ b/resources/src/main/java/org/hl7/ShortNameMaps.java
@@ -16,7 +16,7 @@ public class ShortNameMaps {
   static {
     PAYOR_SHORT_NAME_TO_FULL_NAME.put("cms", "Centers for Medicare and Medicaid Services");
     PAYOR_SHORT_NAME_TO_FULL_NAME.put("uhc", "United Health Care");
-    PAYOR_SHORT_NAME_TO_FULL_NAME.put("medicare", "Medicare");
+    PAYOR_SHORT_NAME_TO_FULL_NAME.put("medicare", "MEDICARE");
   }
 
 }
--- a/resources/src/main/java/org/hl7/ShortNameMaps.java
+++ b/resources/src/main/java/org/hl7/ShortNameMaps.java
@@ -16,6 +16,7 @@ public class ShortNameMaps {
   static {
     PAYOR_SHORT_NAME_TO_FULL_NAME.put("cms", "Centers for Medicare and Medicaid Services");
     PAYOR_SHORT_NAME_TO_FULL_NAME.put("uhc", "United Health Care");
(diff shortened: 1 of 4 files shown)


resources/src/main/java/org/hl7/ShortNameMaps.java
```diff
@@ -16,6 +16,7 @@
   static {
     PAYOR_SHORT_NAME_TO_FULL_NAME.put("cms", "Centers for Medicare and Medicaid Services");
     PAYOR_SHORT_NAME_TO_FULL_NAME.put("uhc", "United Health Care");
+    PAYOR_SHORT_NAME_TO_FULL_NAME.put("medicare", "MEDICARE");
```
plarocque4, line 19: Is Medicare a different payer from CMS?

server/src/main/java/org/hl7/davinci/endpoint/controllers/FhirController.java
```diff
@@ -76,7 +76,7 @@
       // If URL starts with this server's base url, pull out id and search by id
       if (url.startsWith(baseUrl)) {
         String valueSetId = url.split("ValueSet/")[1];
-        FileResource fileResource = fileStore.getFhirResourceById("R4", "valueset", "valueset/" + valueSetId, baseUrl);
+        FileResource fileResource = fileStore.getFhirResourceById("R4", "valueset", valueSetId, baseUrl);
```
plarocque4, line 79: You verified that this still works with all of our previous valuesets and rule sets?

resources/src/main/java/org/hl7/ShortNameMaps.java
yuliumitre: I'm actually not sure.

server/src/main/java/org/hl7/davinci/endpoint/controllers/FhirController.java
yuliumitre: Yes, as our RI actually doesn't use those locally defined value sets. There is only one defined - COPD (maybe for demo purpose?), but it is not used anywhere. This is a path mixup.

resources/src/main/java/org/hl7/ShortNameMaps.java
lpdecelles: "Medicare Advantage" or MA is the plan that CMS offers. You might see Medicare Advantage commercials on TV. So we should remove the MEDICARE line.

lpdecelles: @yuliumitre said she will remove the "Medicare Advantage" line.

PR #230, opened 2021-03-30
Alternative Therapies Support

plarocque4:
Return alternative therapies as suggestions in the cards returned from a CRD CDS Hooks request.
Bump up the versions of several libraries we are using within CRD.


plarocque4:
This change goes with the CDS Library and Request Generator changes to support alternative therapies.

server/src/main/java/org/hl7/davinci/endpoint/cql/r4/CqlExecutionContextBuilder.java
```diff
@@ -38,14 +43,15 @@ public static Context getExecutionContext(CqlRule cqlRule, HashMap<String, Resou
...
     }
 
+    FhirContext fhirContext = FhirContext.forR4();
     Context context = new Context(library);
     context.registerLibraryLoader(libraryLoader);
     context.setExpressionCaching(true);
 
-    BaseFhirDataProvider provider = new DummyFhirDataProvider();
+    R4FhirModelResolver modelResolver = new R4FhirModelResolver();
+    RestFhirRetrieveProvider retrieveProvider = new RestFhirRetrieveProvider(new SearchParameterResolver(fhirContext), fhirContext.newRestfulGenericClient("http://fhirtest.uhn.ca/baseR4"));
```
KeeyanGhoreshi, line 52: is this provider meant to be hard coded?

server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirRequestProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+        .setSystem(alternativeTherapy.getSystem());
+    codingList.add(code);
+    codeableConcept.setCoding(codingList);
+
+    // convert the request into real object type
+    switch (request.fhirType()) {
+      case "DeviceRequest":
+        DeviceRequest deviceRequest = ((DeviceRequest) request).copy();
+        deviceRequest.setCode(codeableConcept);
+        deviceRequest.setId(new IdType()); // clear the ID
```
KeeyanGhoreshi, line 35: I think the Id, Identifier, Meta, and AuthoredOn properties are part of IBaseResource.  It seems like all the case statements do the same thing with these properties, could probably pull these out and do it on the higher level class so it doesn't have to repeat several times.

KeeyanGhoreshi (commented):
Looks pretty good overall

server/src/main/java/org/hl7/davinci/endpoint/cql/r4/CqlExecutionContextBuilder.java
plarocque4: These changes were necessary to get it to work with the newer versions of HAPI I believe. I got this from some examples.

server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirRequestProcessor.java
plarocque4: Only the Id and Meta are part of the base resource (DomainResource) class. I think it'll be too messy to try to. rearrange the code for just those two.

PR #231, opened 2021-04-13
Support encoding base64 cql and embedding it into the FHIR Library.

plarocque4:


plarocque4:
This change goes with the DTR and CDS Library changes to support base64 encoded CQL in the FHIR Library.

KeeyanGhoreshi (commented):
the code looks fine, is there a good way to test it?

plarocque4:
> the code looks fine, is there a good way to test it?

You can turn on the new flag in the configuration and just run through DTR. You will notice in the network traffic that the Library resources now contain base64 encoded CQL. For it to render properly in DTR you will need the DTR repo and cds library repo as well.

PR #232, opened 2021-04-22
Standalone smart app

KeeyanGhoreshi:
To test the standalone smart app, go to `localhost:8090/launch` to open the launch form, input the EHR url and client ID.  Note that you will likely have to register a new client, since the base url for the smart app is `localhost:8090`, not `localhost:3005` like the DTR app.  Once you have a client set up and the client ID and iss in the correct fields, click "launch" and the go through the authentication process.  

On the index page, you can create hooks requests by putting info into the inputs, or by changing the JSON directly.  In order to make a valid hooks request to our CRD server, you need to make sure that the organization field on the coverage resource is recognized.  Use "Centers for Medicare and Medicaid Services" for both organization fields.  The other fields can be changed however you like.  Then click submit, and you should see a card returned on the right.


plarocque4:
A couple of notes on testing this so far.
1) It would be nice to be able to auto hide the editing window for those that don't want to edit the json directly.
2) We need some documentation in the README about how the launch works along with the client information.
3) If all we support right now is CMS, then it should be an option for the coverage organization instead of just a free text box.

plarocque4 (approved):
I like the changes, works great.

PR #236, opened 2021-05-14
CORS changes to support LogicaHealth

plarocque4:


yuliumitre:
It looks good to me.

KeeyanGhoreshi (commented):
looks fine, but we should consider a smoother solution, like env variables or db registration, for after the connectathon.  We shouldn't need to update this list manually any time another endpoint requires access to CRD.

plarocque4:
This has been superseded by the pull request 238.
https://github.com/HL7-DaVinci/CRD/pull/238

PR #237, opened 2021-05-26
Fix for https://jira.mitre.org/browse/DMEERX-967 - Card Source

lpdecelles:
To test just run CRD and look at the card source in Developer Tools.

### DIFF
--- a/server/src/main/java/org/hl7/davinci/endpoint/components/CardBuilder.java
+++ b/server/src/main/java/org/hl7/davinci/endpoint/components/CardBuilder.java
@@ -7,6 +7,7 @@ import java.util.List;
 import org.cdshooks.*;
 import org.hl7.davinci.FhirComponentsT;
 import org.hl7.davinci.endpoint.cdshooks.services.crd.r4.FhirRequestProcessor;
+import org.hl7.davinci.endpoint.config.YamlConfig;
 import org.hl7.fhir.instance.model.api.IBaseResource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -15,7 +16,18 @@ import org.slf4j.LoggerFactory;
  * Convenience methods for working with CDS Hooks cards.
  */
 public class CardBuilder {
+
   static final Logger logger = LoggerFactory.getLogger(CardBuilder.class);
+
+  private YamlConfig myConfig;
+
+  // Default Constructor
+  public CardBuilder() {}
+
+  public CardBuilder(YamlConfig myConfig)
+  {
+    this.myConfig = myConfig;
+  }
 
   public static class CqlResultsForCard {
     private Boolean ruleApplies;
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/hl7/davinci/endpoint/config/YamlConfig.java
```diff
@@ -25,6 +25,8 @@
   private boolean checkPractitionerLocation;
   private boolean appendParamsToSmartLaunchUrl;
   private String hostOrg;
+  private String CDSHooksSourceLabel;
```
plarocque4, line 28: Please rename this to cdsHookSourceLabel, so the capitalization matches the other variables.

server/src/main/resources/application.yml
```diff
@@ -67,3 +67,5 @@ localDb:
 valueSetCachePath: ValueSetCache/
 
 hostOrg: default
+
+CDSHooksSourceLabel: "MITRE Corporation CRD Reference Implementation"
```
plarocque4, line 71: Rename to cdsHooksSourceLabel please.

lpdecelles: Same comment above.

server/src/main/java/org/hl7/davinci/endpoint/config/YamlConfig.java
lpdecelles: According this this link https://google.github.io/styleguide/javaguide.html#s5.3-camel-case. It should look like this: "cDSHookSourceLabel".             Thoughts?

plarocque4: I like cdsHookSourceLabel better personally, cDSHookSourceLabel seems to put a weird emphasis on the DS and not the c otherwise. I could be convinced to go either way though.

server/src/main/java/org/hl7/davinci/endpoint/components/CardBuilder.java
```diff
@@ -15,8 +16,19 @@
  * Convenience methods for working with CDS Hooks cards.
  */
 public class CardBuilder {
+
   static final Logger logger = LoggerFactory.getLogger(CardBuilder.class);
 
+  private YamlConfig myConfig;
```
plarocque4, line 22: I wonder if we should bother passing in the YamlConfig or if we should have it be auto wired like in CdsService.java? Might be good to ask someone that is an expert in java.
`
@Autowired  private 
YamlConfig myConfig;
`

server/src/main/java/org/hl7/davinci/endpoint/config/YamlConfig.java
lpdecelles: I agree, I will switch it to cdsHookSourceLabel.

server/src/main/java/org/hl7/davinci/endpoint/components/CardBuilder.java
lpdecelles: I tried to auto wire YamlConfig. But because that method is static it could not reference/compile it.

plarocque4: That makes sense, I didn't see that it was static. If all the methods that it is used in are static and requires the config to be passed in anyway, do you need to bother having it as a member variable of the CardBuilder class? If not you could remove the constructor as well.

plarocque4:
Can you forward merge from master into this branch so we can get this merge request finished?

KeeyanGhoreshi:
Reopen when this can be finished

PR #238, opened 2021-05-26
Update CORS Security Configuration

plarocque4:
Update CORS Security Configuration to read values from the configuration file or environment variables.

### DIFF
--- a/server/src/main/resources/application.yml
+++ b/server/src/main/resources/application.yml
@@ -25,6 +25,12 @@ server:
 
 checkJwt: false
 
+# CORS Allowed Origins
+corsOrigins:
+  - "http://localhost:8080"
+  - "http://localhost:3000"
+  - "http://localhost:3002"
+
 checkPractitionerLocation: false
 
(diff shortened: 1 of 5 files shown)


server/src/main/resources/application.yml
```diff
@@ -25,6 +25,12 @@ server:
 
 checkJwt: false
 
+# CORS Allowed Origins
+corsOrigins:
+  - "http://localhost:8080"
+  - "http://localhost:3000"
+  - "http://localhost:3002"
+
```
yuliumitre, line 33: Will Logica endpoints be added to the list? I thought that is the reason we wanted to do this story.

plarocque4: Since it is in the configuration I am able to override it with an environment variable. I pushed this same change to the may-2021-connectathon branch and put it on LogicaHealth and was able to set an environment variable and it worked. Having the defaults as just localhost is what we'd likely want to leave in here for someone testing it on their own.

yuliumitre: Sounds good.

yuliumitre (approved):

PR #239, opened 2021-06-07
Fixed SetupGuideForMacOS.md

lpqdao:


yuliumitre (approved):
The changes look good to me. Great work!

SetupGuideForMacOS.md
```diff
@@ -164,16 +166,16 @@ Additionally, you must have credentials (api key) access to the **[Value Set Aut
 
 ### crd-request-generator configs
 
-1. cd `<drlsroot>/dtr` 
-2. In `webpack.config.dev.js`, set `https` to `false`. Your file should look something like this:
+1. cd `<drlsroot>/dtr/bin` 
+2. In `dev`, set `https` to `false`. Your file should look something like this:
```
KeeyanGhoreshi, line 170: this reminds me that https isn't supported by the app right now.  we should probably get that set up.  We can leave this step in here for now, I suppose.

SetupGuideForMacOS.md
```diff
@@ -164,16 +166,16 @@ Additionally, you must have credentials (api key) access to the **[Value Set Aut
 
 ### crd-request-generator configs
```
KeeyanGhoreshi, line 167: This is the header for the DTR section, but it says request generator.  Not your change, but we may as well fix it here while we're at it.

KeeyanGhoreshi: Github won't let me comment further up, but`### dtr configs` above should be the request generator header.  They must've been accidentally swapped.

PR #240, opened 2021-06-09
Swapped crd-request-generator and dtr

lpqdao:


KeeyanGhoreshi:
This should be an extra commit on the original branch.  It works out the same but it's cleaner if everything is in one branch/pull request.

PR #244, opened 2021-08-02
Questionnaire url

KeeyanGhoreshi:
To be used with DTR's questionnaire url branch.  This makes CRD return the fhir/filepath in the appcontext.  It still send the `template`, which is the URI of the questionnaire, but it also send the base URL of the fhir server as `fhirpath` for use by the DTR app.  It also makes `filepath` work again, instead of just being a placeholder.  

These changes will likely need to be reflected in the IG.


plarocque4 (approved):
Looks good

PR #247, opened 2021-09-23
[DMEERX-1043] CRD STU2 - Add ability to annotate order with note based on CRD return

plarocque4:
Changes must be tested and coordinated between CRD, CDS-Library and crd-request-generator all with the same branch name of annotation-suggestion.

Sending a Hospital Bed DeviceRequest for William Oster will return Documentation Required with links to launch into DTR.
Sending a Hospital Bed DeviceRequest for Theodore Roosevelt will return No Prior Authorization required with a link to "Save Update To EHR". When clicked it'll save an updated device request with a note about the prior authorization. You can load the DeviceRequest from the test-ehr to view the note.


KeeyanGhoreshi (approved):
works as expected

PR #255, opened 2021-11-30
Preemptive prior auth

plarocque4:
Add support for preemptive prior auth.

To test you must include the changes for the crd-request-generator in the preemptive-prior-auth branch as well as the CDS-Library in the preemptive-prior-auth branch. Choose Theodor Roosevelt and G0180 Service Request for Medicare-covered home health services under a home health plan of care.

### DIFF
--- a/resources/src/main/java/org/cdshooks/Action.java
+++ b/resources/src/main/java/org/cdshooks/Action.java
@@ -15,6 +15,7 @@ public class Action {
   private TypeEnum type = null;
   private String description = null;
   private IBaseResource resource = null;
+  private String resourceId = null;
 
   private FhirComponentsT fhirComponents;
 
@@ -26,9 +27,7 @@ public class Action {
     this.type = type;
   }
 
-  public String getDescription() {
-    return description;
-  }
+  public String getDescription() { return description; }
 
   public void setDescription(String description) {
     this.description = description;
@@ -36,7 +35,14 @@ public class Action {
 
   public IBaseResource getResource() { return resource; }
 
-  public void setResource(IBaseResource resource) { this.resource = resource; }
+  public void setResource(IBaseResource resource) {
+    this.resource = resource;
+    setResourceId(resource.fhirType() + "/" + resource.getIdElement().getIdPart());
+  }
+
+  public String getResourceId() { return resourceId; }
+
+  public void setResourceId(String resourceId) { this.resourceId = resourceId; }
 
   public FhirComponentsT getFhirComponents() {
     return this.fhirComponents;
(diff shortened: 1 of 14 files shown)


KeeyanGhoreshi (approved):
Looks good to me

yuliumitre:
The PR works well. Have a question about those two buttons after the preemptive prior-auth is granted: the first time I clicked the buttons, it works well. I am wondering if the clicking order matters? As adding prior-auth to Request is using a reference? Does it need to store the ClaimResponse first, then you can add it to the order? 

After the first submit, then clicked the two buttons, those two buttons are disabled. Then I submitted the same request, those two buttons remain disabled although the prios-auth GUID is different. I am wondering if this is a valid use case. If not, it will probably better reenable the buttons after submit?

RobertScalfani (approved):

plarocque4:
> The PR works well. Have a question about those two buttons after the preemptive prior-auth is granted: the first time I clicked the buttons, it works well. I am wondering if the clicking order matters? As adding prior-auth to Request is using a reference? Does it need to store the ClaimResponse first, then you can add it to the order?
> 
> After the first submit, then clicked the two buttons, those two buttons are disabled. Then I submitted the same request, those two buttons remain disabled although the prios-auth GUID is different. I am wondering if this is a valid use case. If not, it will probably better reenable the buttons after submit?

I'll need to fix the request generator to handle these problems. When one of these suggestions is clicked, the other should gray out. They should also both be available as options again after the new CRD request CARD response is received.

yuliumitre:
> > The PR works well. Have a question about those two buttons after the preemptive prior-auth is granted: the first time I clicked the buttons, it works well. I am wondering if the clicking order matters? As adding prior-auth to Request is using a reference? Does it need to store the ClaimResponse first, then you can add it to the order?
> > After the first submit, then clicked the two buttons, those two buttons are disabled. Then I submitted the same request, those two buttons remain disabled although the prios-auth GUID is different. I am wondering if this is a valid use case. If not, it will probably better reenable the buttons after submit?
> 
> I'll need to fix the request generator to handle these problems. When one of these suggestions is clicked, the other should gray out. They should also both be available as options again after the new CRD request CARD response is received.

Sounds good. Go ahead and merge this PR. Thanks.

yuliumitre (approved):

plarocque4:
> > > The PR works well. Have a question about those two buttons after the preemptive prior-auth is granted: the first time I clicked the buttons, it works well. I am wondering if the clicking order matters? As adding prior-auth to Request is using a reference? Does it need to store the ClaimResponse first, then you can add it to the order?
> > > After the first submit, then clicked the two buttons, those two buttons are disabled. Then I submitted the same request, those two buttons remain disabled although the prios-auth GUID is different. I am wondering if this is a valid use case. If not, it will probably better reenable the buttons after submit?
> > 
> > 
> > I'll need to fix the request generator to handle these problems. When one of these suggestions is clicked, the other should gray out. They should also both be available as options again after the new CRD request CARD response is received.
> 
> Sounds good. Go ahead and merge this PR. Thanks.

I made some changes to this branch to add in missing values in the Action, Suggestion, and Card. Please review these new changes!

yuliumitre (approved):
It works well. I can see the flag is added to the card.

PR #256, opened 2021-12-02
Adform - CQL, Multiple Questions, SaveToEHR support

RobertScalfani:
This PR adds support for CQL, Multiple Questions sent at once for Adforms, and SaveToEHR support so that questionniares can be resumed at any point in the process.
Also improves and refactors the stability and design of the adform next-question tree.

This PR also introduces the Adaptive form next question tree implementation that was created on the `adForm-next-question-tree` branch.

To be used with:

- CDS-Library/`adform-cql-multiple-questions`
- dtr/`cql`


yuliumitre:
main/java/org/hl7/davinci/endpoint/controllers/Questions-HomeOxygenTherapyAdditional.json 

This file should not be checked in, right? As we move the Questions file to CDS-Library.

server/src/main/java/org/hl7/davinci/endpoint/files/SubQuestionnaireProcessor.java
```diff
@@ -76,6 +76,10 @@ private void processItemList(List<QuestionnaireItemComponent> itemList, FileStor
     for (int i = 0; i < itemList.size();) {
       List<QuestionnaireItemComponent> returnedItemList = 
         processItem(itemList.get(i), fileStore, baseUrl, containedList, extensionList);
+
+      // if(itemList.get(i).getItem().size() > 0){
+      //   this.processItemList(itemList.get(i).getItem(), fileStore, baseUrl, containedList, extensionList);
+      // }
```
yuliumitre, line 82: Commented out code needs to be cleaned up if not needed anymore.

server/src/main/java/org/hl7/davinci/endpoint/controllers/QuestionnaireController.java
```diff
@@ -0,0 +1,427 @@
...
+
+            String questionnaireId = ((Reference) inputQuestionnaireResponse.getExtensionByUrl("http://hl7.org/fhir/StructureDefinition/contained-id").getValue()).getReference();
+            System.out.println("Input Questionnaire: " + questionnaireId);
+
+            if (inputQuestionnaireFromRequest != null) {
+
+                if(!questionnaireTrees.containsKey(questionnaireId)){
+                    // If there is not already a tree that matches the requested questionnaire id, build it.
+                    // Import the requested CDS-Library Questionnaire.
+                    Questionnaire cdsQuestionnaire = QuestionnaireController.importCdsQuestionnaire(request, parser, fileStore, "Questions-HomeOxygenTherapyAdditionalAdaptive.json");
```
yuliumitre, line 338: Although we have only one adaptive form, it will probably better to retrieve the Questions-*.json file from the contained Questionnaire id instead of hard coding it.

server/src/main/java/org/hl7/davinci/endpoint/controllers/QuestionnaireController.java
```diff
@@ -0,0 +1,427 @@
...
+            if(pullFromResources){
+                // Resource is pulled from the file store as a resource.
+                String baseUrl = Utils.getApplicationBaseUrl(request).toString() + "/";
+                FileResource fileResource = fileStore.getFhirResourceById("r4", "questionnaire", questionnaireId, baseUrl);
+                org.springframework.core.io.Resource resource = fileResource.getResource();
+                InputStream resourceStream = resource.getInputStream();
+                cdsQuestionnaire = (Questionnaire) parser.parseResource(resourceStream);
+            } else {
+                // File is pulled from the file store as a file.
+                FileResource fileResource = fileStore.getFile("HomeOxygenTherapy", questionnaireId, "R4", false);
```
yuliumitre, line 396: Same comments here about hard coding file.

server/src/main/java/org/hl7/davinci/endpoint/controllers/QuestionnaireController.java
```diff
@@ -0,0 +1,427 @@
...
+
+                QuestionnaireResponseItemComponent currentQuestionResponse = currentQuestionResponses.get(0);
+                QuestionnaireResponseItemAnswerComponent currentQuestionAnswer = currentQuestionResponse.getAnswerFirstRep();
+
+
+                // With the currrent question answer in hand, extract the next question.
+                Type value = currentQuestionAnswer.getValue();
+                String response;
+                // Not a fan of this, but not sure how else to determine the value based on different possible input types.
+                if(value instanceof StringType){
```
KeeyanGhoreshi, line 189: you could use `currentQuestionAnswer.hasValueStringType()` here instead of checking the instance yourself, though I'll note that this function does the same thing but also includes a null check.   Makes it a bit more readable though.

KeeyanGhoreshi (commented):
Tested and it works, just need to remove some of the hard-coded value.

server/src/main/java/org/hl7/davinci/endpoint/controllers/QuestionnaireController.java
RobertScalfani: Thanks, updated it with this suggestion!

yuliumitre (approved):
Looks good to me.

KeeyanGhoreshi (approved):
Will wait to merge until other components are ready

PR #265, opened 2022-02-18
Query Batch Request

RobertScalfani:
This PR adds Query Batch Request functionality that attempts to backfill required data that may have been missed by either the Prefetch or PrefetchHydrator.
This implementation successfully pulls the insurance:Coverage resources that do not support _includes for insurance.
It also supports nested Query Batch Requests (i.e. pulling Organization/Practitioner/etc. from PractitionerRole)

It also adds .toString() implementations for many objects from when I was figuring things out. I figure they can be useful going forward, let me know if I should remove them.


server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirBundleProcessor.java
```diff
@@ -175,7 +175,8 @@ public void processOrderSelectMedicationStatements() {
       throw e;
     } catch (Exception e) {
       // catch all remaining exceptions
-      System.out.println(e);
+      System.out.println("Create criteria list exception: ");
+      e.printStackTrace();
```
plarocque4, line 179: Do we want to leave the stack trace in?

server/src/main/java/org/hl7/davinci/endpoint/components/QueryBatchRequest.java
```diff
@@ -0,0 +1,246 @@
...
+public class QueryBatchRequest {
+
+  private static final Logger logger = LoggerFactory.getLogger(QueryBatchRequest.class);
+  private static final String PRACTIONER_ROLE = "PractitionerRole";
+
+  // private final CdsService<?> cdsService;
+  private final CdsRequest<CrdPrefetch, ServiceContext> cdsRequest;
+  private final FhirComponentsT fhirComponents;
+
+  public QueryBatchRequest(CdsService<?> cdsService, CdsRequest<?, ?> cdsRequest, FhirComponentsT fhirComponents) {
```
plarocque4, line 51: It looks like you don't need to pass in the cdsService since you aren't using it. You can also remove the commented out lines above and below.

server/src/main/java/org/hl7/davinci/endpoint/components/QueryBatchRequest.java
```diff
@@ -0,0 +1,246 @@
...
+    // A Query Batch Request is used to try and populate missing fields that the
```
plarocque4, line 52: Move the comments to above the class definition.

server/src/main/java/org/hl7/davinci/endpoint/components/QueryBatchRequest.java
```diff
@@ -0,0 +1,246 @@
...
+   * querying and adding the attributes they missed.
+   * Approach:
+   * 1. Pull the IDs of the required references from the request object's draft
+   * orders.
+   * 2. See which of those values are missing from the current CRD response.
+   * 3. Build the Query Batch JSON request using
+   * http://build.fhir.org/ig/HL7/davinci-crd/hooks.html#fhir-resource-access
+   * 4. Populate the CRD response with the values from the Query Batch response.
+   */
+  public void performQueryBatchRequest() {
```
plarocque4, line 71: Is it better to pass in the cdsRequest here and pass it to everywhere needed?

server/src/main/java/org/hl7/davinci/endpoint/components/QueryBatchRequest.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+    String fullUrl = fhirBase + "/";
+
+    String token = null;
+    if (cdsRequest.getFhirAuthorization() != null) {
+      token = cdsRequest.getFhirAuthorization().getAccessToken();
+    }
+
+    RestTemplate restTemplate = new RestTemplate();
+    HttpHeaders headers = new HttpHeaders();
```
plarocque4, line 166: Is this the same way that queries are made to the EHR FHIR server for prefetch?

RobertScalfani: I went ahead and moved executeFhirQuery() to FhireRequestProcessor.java. In doing this, I also deleted executeFhirQuery() in PrefetchHydrator.java since they follow the same format so now the hydrator and query batch use the same executeFhirQuery() method.

resources/src/main/java/org/hl7/davinci/r4/Utilities.java
```diff
@@ -169,10 +169,12 @@ public static PractitionerRoleInfo getPractitionerRoleInfo(
   public static List<Organization> getPayors(List<Coverage> coverages) {
     List<Organization> payors = new ArrayList<>();
     for (Coverage coverage: coverages){
-      for (Reference ref: coverage.getPayor()){
-        Organization organization = (Organization) ref.getResource();
-        if (organization != null) {
-          payors.add(organization);
+      if(coverage != null){
```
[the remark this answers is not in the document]
RobertScalfani: Removed that check.
Added line 153 in FhirBundleProcessor.java that filters null values that may not have resolved to address that issue.

server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/CdsService.java
```diff
@@ -151,16 +155,27 @@ public CdsResponse handleRequest(@Valid @RequestBody requestTypeT request, URL a
     // hydrated
     requestLog.advanceTimeline(requestService);
 
-    // logger.info("***** ***** request from requestLog: "+requestLog.toString() );
+    // Attempt a Query Batch Request to backfill missing attributes.
+    if(myConfig.isQueryBatchRequest()){
+      QueryBatchRequest batchRequest = new QueryBatchRequest(this.fhirComponents);
+      batchRequest.performQueryBatchRequest((CdsRequest<CrdPrefetch, ?>) request);
```
KeeyanGhoreshi, line 161: this is more of an observation than a comment for change, but CrdPrefetch here breaks pattern with the way CdsService defines requestTypeT and then implementing classes define it.  The result is that CdsService is deciding preemptively what class to cast the request, which it shouldn't be doing.  This doesn't cause any problems because currently the only thing the wildcard gets cast to is `org.hl7.davinci.r4.crdhook.CrdPrefetch`.  

The options here are to reform the structure and remove this wildcard so that the class only takes CrdPrefetch or redoing the QueryBatchRequest function to operate like how the PrefetchHydrator operates, depending on whether we'd expect the Prefetch to have another class associated with it.  This hasn't been the case since we removed R3 from the repo.

plarocque4 (approved):
Looks good

PR #267, opened 2022-03-14
Update the annotations to include an extension with discrete values.

plarocque4:
Update the annotations to include an extension with discrete values.
Add annotations to more cards returned.

### DIFF
--- a/server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirRequestProcessor.java
+++ b/server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirRequestProcessor.java
@@ -1,7 +1,6 @@
 package org.hl7.davinci.endpoint.cdshooks.services.crd.r4;
 
 import org.cdshooks.AlternativeTherapy;
-import org.hl7.fhir.instance.model.api.IBase;
 import org.hl7.fhir.instance.model.api.IBaseResource;
 import org.hl7.fhir.r4.model.*;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirRequestProcessor.java
```diff
@@ -164,4 +163,57 @@ public static IBaseResource addSupportingInfoToRequest(IBaseResource request, Re
...
+        break;
+      case "MedicationRequest":
+        MedicationRequest medicationRequest = ((MedicationRequest) request).copy();
+        coverage = medicationRequest.getInsurance().get(0);
+        break;
+      case "ServiceRequest":
+        ServiceRequest serviceRequest = ((ServiceRequest) request).copy();
+        coverage = serviceRequest.getInsurance().get(0);
+        break;
+      case "MedicationDispense":
```
KeeyanGhoreshi, line 206: Is dispense no longer supported?  One of the patients has Azathioprine as a MedicationDispense, and the request works on master, but this change throws an error instead.

lpdecelles:
MedicationDispense should still be supported.

server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirRequestProcessor.java
plarocque4: Good catch, I am pushing up a fix for this. The problem has to do with the way that I'm getting the coverage from the request. The MedicationDispense does not have coverage included.

server/src/main/java/org/hl7/davinci/endpoint/components/CardBuilder.java
```diff
@@ -311,33 +311,39 @@ public static Suggestion createSuggestionWithNote(Card card,
...
+      date.setUrl("date")
+          .setValue(new DateType().setValue(now));
+      extension.addExtension(date);
+      Extension identifier = new Extension();
+      String id = UUID.randomUUID().toString();
+      identifier.setUrl("identifier")
+          .setValue(new StringType(id));
+      extension.addExtension(identifier);
+      resource = FhirRequestProcessor.addExtensionToRequest(resource, extension);
+    } catch (RuntimeException e) {
```
KeeyanGhoreshi, line 344: This feels a bit broad for the exception here, is the only problem that the coverage is not available?  If so, can we check for that in particular?

plarocque4: I added a more specific exception.

KeeyanGhoreshi (approved):

PR #271, opened 2022-04-25
Questionnaire-package to support DTR relaunch

plarocque4:
Add a Questionnaire-package operation to the Payer FHIR Server portion of CRD. This supports the DTR relaunch detailed in the DTR STU1.1 IG. Please note that this is not part of the CRD IG and is only included in CRD because CRD is also acting as the Payer FHIR Server to support DTR.

### DIFF
--- a/resources/src/main/java/org/hl7/davinci/r4/Utilities.java
+++ b/resources/src/main/java/org/hl7/davinci/r4/Utilities.java
@@ -4,6 +4,7 @@ import java.util.*;
 
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.parser.IParser;
+
 import org.hl7.davinci.*;
 import org.hl7.fhir.instance.model.api.IBaseResource;
 import org.hl7.fhir.r4.model.*;
@@ -169,10 +170,12 @@ public class Utilities {
   public static List<Organization> getPayors(List<Coverage> coverages) {
     List<Organization> payors = new ArrayList<>();
     for (Coverage coverage: coverages){
-      for (Reference ref: coverage.getPayor()){
-        Organization organization = (Organization) ref.getResource();
-        if (organization != null) {
-          payors.add(organization);
+      if (coverage != null) {
+        for (Reference ref: coverage.getPayor()){
+          Organization organization = (Organization) ref.getResource();
+          if (organization != null) {
+            payors.add(organization);
+          }
         }
       }
     }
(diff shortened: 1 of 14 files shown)


yuliumitre:
@plarocque4 I got 500 with null pointer.  Will send you the postman collection I used. Let me know if I missed anything. Thanks. 

2022-04-25 15:29:21.002 ERROR 46958 --- [nio-8090-exec-3] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is java.lang.NullPointerException] with root cause

java.lang.NullPointerException: null
        at org.hl7.davinci.endpoint.fhir.r4.QuestionnairePackageOperation.getResource(QuestionnairePackageOperation.java:148) ~[main/:na]
        at org.hl7.davinci.endpoint.fhir.r4.QuestionnairePackageOperation.execute(QuestionnairePackageOperation.java:66) ~[main/:na]
        at org.hl7.davinci.endpoint.controllers.FhirController.questionnaireForOrderOperation(FhirController.java:240) ~[main/:na]

plarocque4:
> Thanks Pat for clarifying the payload format. The operation works as advertised 👍
> 
> One question I have is about the valueset: based on the IG, the $questionnaire-package should return the value set together with rest of the resources. Otherwise, DTR needs to call valueset-expand operation? Is there a separate ticket for that?

This is a good question. I thought that since it was retrieving a bunch of ValueSets that it had all of them contained. Are there any missing? Would you expect something different in the output? If so, I will need to make some changes.

yuliumitre:
> Thanks Pat for clarifying the payload format. The operation works as advertised 👍
> 
> One question I have is about the valueset: based on the IG, the $questionnaire-package should return the value set together with rest of the resources. Otherwise, DTR needs to call valueset-expand operation? Is there a separate ticket for that?



> > Thanks Pat for clarifying the payload format. The operation works as advertised 👍
> > One question I have is about the valueset: based on the IG, the $questionnaire-package should return the value set together with rest of the resources. Otherwise, DTR needs to call valueset-expand operation? Is there a separate ticket for that?
> 
> This is a good question. I thought that since it was retrieving a bunch of ValueSets that it had all of them contained. Are there any missing? Would you expect something different in the output? If so, I will need to make some changes.


According to the Lloyd's JIRA ticket with comments for detailed description of the operation: https://jira.hl7.org/browse/FHIR-36532 
It indicates the output should include expanded valueset if the server is able to expand the value set. In addition, it says the value set and library shouldn't be repeated in the out put even they appear in multiple questionnaires. The payload will be really big with expanded value set but it does save additional network call to expand the value set. We can ask Lloyd to further clarify the requirements.

yuliumitre:
Noticed some value sets are embedded as "contained" in the Questionnaire. Checked the original Questionnaire doesn't have this section. Wondering why we want to embed VS? It will be hard to share the same VS if they are used in different Questionnaires in the same bundle. 
Screen Shot 2022-04-29 at 2 31 25 PM

plarocque4:
> Noticed some value sets are embedded as "contained" in the Questionnaire. Checked the original Questionnaire doesn't have this section. Wondering why we want to embed VS? It will be hard to share the same VS if they are used in different Questionnaires in the same bundle.
> Screen Shot 2022-04-29 at 2 31 25 PM

I didn't add any code to handle this any differently, I called code that already existed that pulled the Questionnaire and had assumed that they were embedded in the regular Questionnaire as well. I'll take a look into this a little more though.

yuliumitre (approved):
It works well as specified.

PR #272, opened 2022-05-03
Fix issues created recently with GitHub file retrieval. 

plarocque4:
Make FHIR version all caps and make sure to read the stream only once when converting CQL to ELM.


zacharyrobin (approved):
Ran locally, no issues on my end

PR #273, opened 2022-05-13
STU1.1 Linking Cards to Requests

plarocque4:
Add the davinci-associated-resource extension to the card containing a reference to the resource that the card is in response to.

To test, look at the output card and verify that there is an extension with a reference to the associated resource.

### DIFF
--- a/resources/src/main/java/org/cdshooks/Card.java
+++ b/resources/src/main/java/org/cdshooks/Card.java
@@ -16,6 +16,7 @@ public class Card {
   private List<Suggestion> suggestions = null;
   private SelectionBehaviorEnum selectionBehavior = null;
   private List<Link> links = null;
+  private CardExtension extension = null;
 
   public Card() { this.uuid = UUID.randomUUID().toString(); }
 
@@ -99,6 +100,14 @@ public class Card {
 
   public void setLinks(List<Link> links) {
     this.links = links;
+  }
+
+  public CardExtension getExtension() {
+    return this.extension;
+  }
+
+  public void setExtension(CardExtension extension) {
+    this.extension = extension;
   }
 
   public enum IndicatorEnum {
--- a/resources/src/main/java/org/cdshooks/CardExtension.java
+++ b/resources/src/main/java/org/cdshooks/CardExtension.java
@@ -1,0 +1,32 @@
+package org.cdshooks;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CardExtension {
+  
+  @JsonProperty("davinci-associated-resource")
+  private List<String> associatedResources;
+
+  public CardExtension() {
+    associatedResources = new ArrayList<>();
+  }
+
+  public CardExtension(List<String> configurationOptions) {
+    this.associatedResources = configurationOptions;
+  }
+
+  public List<String> getAssociatedResources() {
+    return associatedResources;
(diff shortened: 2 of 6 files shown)


yuliumitre:
@plarocque4  I am seeing the following error related to MedicationDispense. 

The card extension works as specified in the story. I am trying to test the Immunosuppress drug to see the link with card, the cds-service is throwing 500. 

The step to reproduce:
1. Select "Theodor Roosevelt" and "azathioprine 50 MG Oral"
2. Submit

2022-05-19 09:46:39.619 ERROR 21749 --- [nio-8090-exec-4] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is java.lang.ClassCastException: class org.hl7.fhir.r4.model.MedicationRequest cannot be cast to class org.hl7.fhir.r4.model.MedicationDispense (org.hl7.fhir.r4.model.MedicationRequest and org.hl7.fhir.r4.model.MedicationDispense are in unnamed module of loader 'app')] with root cause

java.lang.ClassCastException: class org.hl7.fhir.r4.model.MedicationRequest cannot be cast to class org.hl7.fhir.r4.model.MedicationDispense (org.hl7.fhir.r4.model.MedicationRequest and org.hl7.fhir.r4.model.MedicationDispense are in unnamed module of loader 'app')
        at org.hl7.davinci.endpoint.cdshooks.services.crd.r4.FhirRequestProcessor.addInsuranceAndSubject(FhirRequestProcessor.java:352) ~[main/:na]
        at org.hl7.davinci.endpoint.components.QueryBatchRequest.performQueryBatchRequest(QueryBatchRequest.java:109) ~[main/:na]
        at org.hl7.davinci.endpoint.cdshooks.services.crd.r4.OrderSignService.attempQueryBatchRequest(OrderSignService.java:230) ~[main/:na]
        at org.hl7.davinci.endpoint.cdshooks.services.crd.r4.OrderSignService.attempQueryBatchRequest(OrderSignService.java:35) ~[main/:na]
        at org.hl7.davinci.endpoint.cdshooks.services.crd.CdsService.handleRequest(CdsService.java:162) ~[main/:na]
        at org.hl7.davinci.endpoint.controllers.r4.CdsHooksController

yuliumitre:
BTW, got the same error with MedicationRequest

plarocque4:
> In general, all the cards returned by the CDS service should be for the same "order". I am wondering what is the use case of the linking cards to requests. Does the CDS client lose the context after it sends the request?

Here is there relevant portion of the IG this is discussed:
http://build.fhir.org/ig/HL7/davinci-crd/hooks.html#linking-cards-to-requests
There could be multiple orders in the same CRD request. The reference implementation currently just looks at the first one, but if there are more than one it is useful to know which order the card is in reference to.

yuliumitre:
> > In general, all the cards returned by the CDS service should be for the same "order". I am wondering what is the use case of the linking cards to requests. Does the CDS client lose the context after it sends the request?
> 
> Here is there relevant portion of the IG this is discussed: http://build.fhir.org/ig/HL7/davinci-crd/hooks.html#linking-cards-to-requests There could be multiple orders in the same CRD request. The reference implementation currently just looks at the first one, but if there are more than one it is useful to know which order the card is in reference to.

Great, this is helpful. Thanks.

yuliumitre (approved):

PR #275, opened 2022-06-17
Dmeerx 1150 Refactor Next Question Endpoint

yuliumitre:
This PR moves the $next-question to FHIRController. 
To test:
1. Switch DTR repo to branch DMEERX-1150
2. Follow the usual request generator, submit to CRD, then launch the adaptive form to launch DTR

### DIFF
--- a/server/src/main/java/org/hl7/davinci/endpoint/fhir/r4/QuestionnaireNextQuestionOperation.java
+++ b/server/src/main/java/org/hl7/davinci/endpoint/fhir/r4/QuestionnaireNextQuestionOperation.java
@@ -1,5 +1,6 @@
-package org.hl7.davinci.endpoint.controllers;
+package org.hl7.davinci.endpoint.fhir.r4;
 
+import org.hl7.davinci.endpoint.files.FileStore;
 import org.hl7.davinci.endpoint.Application;
 import org.hl7.davinci.endpoint.files.FileResource;
 import org.hl7.davinci.endpoint.files.FileStore;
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/hl7/davinci/endpoint/fhir/r4/QuestionnaireNextQuestionOperation.java
```diff
@@ -267,29 +265,14 @@ private boolean isLeafNode() {
...
-    /**
+    
+     /**
      * Returns the next question based on the request.
      * @param body
      * @param request
      * @return
      */
-    private ResponseEntity<String> getNextQuestionOperation(String body, HttpServletRequest request) {
+    public ResponseEntity<String> getNextQuestionOperation(String body, HttpServletRequest request) {
```
plarocque4, line 275: How do you feel about renaming this method to execute()?
The class is already the operation, so maybe you can execute it with a call like was done with the QuesionnaireForOrderOperation.

plarocque4: Which request do I submit to trigger the next-question operation?

yuliumitre: Yeah, sure, execute() sounds better and is consistent with other operations' method.

yuliumitre:
> Which request do I submit to trigger the next-question operation?

I will send you the postman collection via email.

plarocque4 (approved):

PR #276, opened 2022-07-15
Add Support For Multiple Prefetch and Updated Prefetch Templates

RobertScalfani:
This PR adds support for accepting multiple prefetch inputs from the request generator. It now meets the spec requirement of prefetch with the format: prefetch:{{deviceRequestBundle: ...},{coverageBundle: ...}, ...}. It supports coverages as a separate bundle in the prefetch.

Updates were also made to the FhirBundleProcessor (in processDeviceRequests(), processServiceRequests(), etc) so that they now properly extract the patient resource itself from the prefetch, not from the request's references. It also implemented structural support for using the coverage bundle, though that will need to be expanded in future to better utilize the coverage resources.

This PR also updated all of the prefetch templates to the newest versions with support for coverage bundle prefethes.
It updated the Prefetch Hydrator to support multiple prefetch and coverage. The necessity for this PR arose from adding a separate Coverage prefetch template outside of the main templates so that coverage can be pulled for all types.
The templates can be viewed at:
https://build.fhir.org/ig/HL7/davinci-crd/hooks.html#prefetch

This PR must be used with the crd-request-generator branch prefetch-tempates. Note the misspelled 'L' in 'tempates'... sorry.
This PR is in conjunction with https://github.com/HL7-DaVinci/crd-request-generator/pull/62 and both must be merged at the same time.

A few notes:
Had to change Coverage?member={{context.patient}} to Coverage?_member={{context.patientId}}
There is a duplicate MedicationRequest template. I assume one should be MedicationDispense, but we need to confirm this with Lloyd.
Some of the templates do not have the coverage as a template element even though in the old version it did. Is this intentional?

### DIFF
--- a/resources/src/main/java/org/hl7/davinci/PrefetchTemplateElement.java
+++ b/resources/src/main/java/org/hl7/davinci/PrefetchTemplateElement.java
@@ -1,10 +1,12 @@
 package org.hl7.davinci;
 
+import org.hl7.fhir.instance.model.api.IBaseResource;
+
 public class PrefetchTemplateElement {
 
   private String key;
   private String query;
-  private Class returnType;
+  private Class<? extends IBaseResource> returnType;
 
   /**
    * Describes a prefetch element. Note that the key and query are exposed in the service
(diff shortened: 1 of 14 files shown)


plarocque4 (commented):
Might be nice to pair down some of the logging, specifically the dumping of the entire prefetch.

RobertScalfani:
> Might be nice to pair down some of the logging, specifically the dumping of the entire prefetch.

resolved

plarocque4 (approved):

PR #278, opened 2022-08-09
Added support for passing coverage to CQL rules

RobertScalfani:
This PR adds support for passing coverages to the CQL rules for the rules to use to determine coverage eligibility.

To test: William Oster has Medicare Part A and Teddy Roosevelt has Part B. The example CQL rule adds a condition to HomeOxygenTherapy'sRULE_APPLIES that the coverage must be Medicare Part A. Thus, for William Oster Home Oxygen Therapy should return the expected cards but for Teddy Roosevelt it should return "No Documentation Rules Found"

Use with branches:
- test-ehr: coverage-rules https://github.com/HL7-DaVinci/test-ehr/pull/65
- CDS-Library: coverage rules https://github.com/HL7-DaVinci/CDS-Library/pull/92

### DIFF
--- a/server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirBundleProcessor.java
+++ b/server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirBundleProcessor.java
@@ -242,6 +242,7 @@ public class FhirBundleProcessor {
     HashMap<String, Resource> cqlParams = new HashMap<>();
     cqlParams.put("Patient", patient);
     if(coverages.size() > 0) {
+      // In this case, we're just using the first coverage. But in a payer implementation, it should find and use the correct relevant coverage(s).
       cqlParams.put("Coverage", coverages.get(0));
     }
     cqlParams.put(requestType, request);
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/hl7/davinci/endpoint/cdshooks/services/crd/r4/FhirBundleProcessor.java
```diff
@@ -234,10 +237,13 @@ private List<CoverageRequirementRuleCriteria> createCriteriaList(CodeableConcept
     }
   }
 
-  private void buildExecutionContexts(List<CoverageRequirementRuleCriteria> criteriaList, Patient patient, String requestType, DomainResource request) {
+  private void buildExecutionContexts(List<CoverageRequirementRuleCriteria> criteriaList, Patient patient, List<Coverage> coverages, String requestType, DomainResource request) {
     System.out.println("buildExecutionContexts::PATIENT: " + patient);
     HashMap<String, Resource> cqlParams = new HashMap<>();
     cqlParams.put("Patient", patient);
+    if(coverages.size() > 0) {
```
plarocque4, line 244: We should probably put a comment in here about how we are just using the first coverage found, but a complete implementation at a payer would likely find the correct coverage and not just assume it is the first one.

plarocque4 (approved):
Looks good.

PR #281, opened 2022-09-02
Query Batch Request Bug Fixes

RobertScalfani:
This PR fixes two bugs with the Query Batch Request:

1. If there were any resources missing that should have been prefetched, the QBR would add a duplicate of the main request resources.
2. The QBR would only look at the first element of the draft orders, meaning that the QBR's requests would be flattened into the first prefetch bundle, rather than being inserted into the correct associated bundles.


plarocque4 (approved):
Works well, and looks clean.

